PR #204, opened 2021-02-19
`plugin update` and `plugin install` should not accept no plugins in args. Closes #199

### LINKED ISSUE #199: `plugin update` and `plugin install` should not accept no plugins in args
`steampipe plugin install` should throw an error pointing out that no plugins were specified

`steampipe plugin update` should throw if the `--all` flag is not provided

binaek89:
Screenshot 2021-02-22 at 10 53 16 PM
Screenshot 2021-02-22 at 10 56 04 PM

@e-gineer @johnsmyth outputs when no plugins are provided.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -206,17 +206,9 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	installSkipped := []string{}
 
 	if len(plugins) == 0 {
-		utils.ShowError(fmt.Errorf(`you need to provide at least one plugin to install.
-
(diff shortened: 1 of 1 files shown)


e-gineer:
Comments:
* Can / should we use the exact output from the help of the command? This seems custom?
* If custom ... should we include the multi-plugin example? Seems complex / not necessary.

binaek89:
@e-gineer @johnsmyth here are some updated screenshots. Removed the custom text and have printed out the `help` text along with an `Error:` at the top. 

Install Output
Update Output

PR #212, opened 2021-02-22
global pluralisation utility. Closes #210

### LINKED ISSUE #210: global pluralisation utility

binaek89:


cmd/plugin.go
```diff
@@ -244,11 +244,24 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	}
 
 	if len(installSkipped) > 0 {
-		fmt.Println("Skipped the following plugins, since they are already installed:")
+		fmt.Printf(
+			"Skipped the following %s, since %s %s already installed:\n",
+			utils.Pluralize("plugin", len(installSkipped)),
+			utils.Pluralize("it", len(installSkipped)),
```
kaidaguerre, line 250: hmm I wonder if we should just define rules for `it is` and `this plugin` - seeing as we now have our own steampipe specific pluralize library - we may as well add some useful phrases

kaidaguerre (changes requested):

PR #214, opened 2021-02-22
Launch failure when using Root. Closes #167

### LINKED ISSUE #167: Launch failure when using Root
Hey guys, ran into a failure when trying to launch as root within WSL (Postgres failed to launch...). Switching back to a non-priv user seems to have fixed the issue. I believe the issue comes from trying to launch the postgres as root and not running the service as a different user, here's what I get when trying to run it directly:

```bash
root@Zindows:~/.steampipe/db/12.1.0/postgres/bin# ./postgres
"root" execution of the PostgreSQL server is not permitted.
The server must be started under an unprivileged user ID to prevent
possible system security compromise.  See the documentation for
more information on how to properly start the server.
```

binaek89:


binaek89:
Screenshot 2021-02-22 at 10 34 45 PM

@e-gineer @johnsmyth  this is what we are showing when `steampipe` is run as a super-user.

main.go
```diff
@@ -48,3 +52,69 @@ func createLogger() {
...
+	 * worth; we'll just expend the effort to check for it.)
+	 */
+
+	if os.Geteuid() != os.Getuid() {
+		utils.ShowError(fmt.Errorf(`%s: real and effective user IDs must match`, constants.Bold("steampipe")))
+		os.Exit(-1)
+	}
+}
+
+// static void
```
kaidaguerre, line 82: can we delete this?

main.go
```diff
@@ -48,3 +52,69 @@ func createLogger() {
...
+	/*
+	 * Also make sure that real and effective uids are the same. Executing as
+	 * a setuid program from a root shell is a security hole, since on many
+	 * platforms a nefarious subroutine could setuid back to root if real uid
+	 * is root.  (Since nobody actually uses postgres as a setuid program,
+	 * trying to actively fix this situation seems more trouble than it's
+	 * worth; we'll just expend the effort to check for it.)
+	 */
+
+	if os.Geteuid() != os.Getuid() {
```
kaidaguerre, line 76: if this is needed, can you get this error signed off by John/Nathan as well

kaidaguerre (changes requested):

johnsmyth (approved):

binaek89:
@johnsmyth @e-gineer @kaidaguerre 

there's another error string which is shown when the `real` userID is not equal to the `effective` userID:

`steampipe: real and effective user IDs must match`

The idea is to replicate the behaviour as is laid out in `postgresql`, so that we can avoid the scenario where `steampipe` starts up, but the underlying `postgresql` refuses to startup.

e-gineer:
Steampipe should not be in bold for the "root" error message above?

Please share exact output for the real / effective error for review.

binaek89:
> Steampipe should not be in bold for the "root" error message above?

👍  I will remove the **bold**

> 
> Please share exact output for the real / effective error for review.

The exact output from `postgres` is 
`postgres: real and effective user IDs must match`

Ours is pretty much an exact copy (except the bold part)

binaek89:
@e-gineer 

update:

I removed the `steampipe:` from the error, since we don't use it anywhere else.

Attaching a screenshot of the actual output

Screenshot 2021-02-24 at 2 14 46 PM

PR #221, opened 2021-02-23
cap column width to available cols in screen. Closes #155

### LINKED ISSUE #155: Cosmetic Issue: The description column is too wide.
When doing a `.inspect` against a table, the `Description` field expands to the full width of the open terminal. This works well if the text in the `Description` column is large, but for tables where all the descriptions are short, the result looks a bit strange. 

image

Suggest that the width of the `DESCRIPTION` column is the `lesser` of [longest description column text, available terminal width]

binaek89:


binaek89:
Screenshot 2021-02-23 at 5 18 48 PM

@johnsmyth @kaidaguerre @e-gineer can you take a look at the final output?

PR #225, opened 2021-02-23
set search_path to `steampipe` user. Closes #218

### LINKED ISSUE #218: Connections must be fully qualified with their schema when querying via service mode
## Issue

If you are querying Steampipe via service mode (e.g. in your own data client, rather than interactively in the CLI) and try to use a relation in a schema that exists, you have to fully-qualify that relation with its schema name.

## Expected outcome

If you have a schema called `aws` and have a table `aws_iam_user` in it, I should be able to query `select * from aws_iam_user` and the correct schema be found via the search path.

## Actual outcome

`select * from aws_iam_user` gives this error:

```
ERROR:  relation "aws_iam_user" does not exist
LINE 1: select * from aws_iam_user;
                      ^
SQL state: 42P01
Character: 15
```

...instead you have to fully-qualify the table with its schema:

`select * from aws.aws_iam_user;`

## Env

CLI: `0.2.0`
OS: `macOS Big Sur 11.2.1`
Terminal: `iTerm 3.4.4` / `zsh`

binaek89:


binaek89:
PgCLI:
Screenshot 2021-02-24 at 3 19 53 PM

psql:
Screenshot 2021-02-24 at 3 21 01 PM

PR #228, opened 2021-02-23
`spinner` should not come up in non-interactive query mode. Closes #227

### LINKED ISSUE #227: `spinner` should not come up in non-interactive query mode

binaek89:


db/interactive_client.go
```diff
@@ -205,7 +205,7 @@ func (c *InteractiveClient) executor(line string, resultsStreamer *ResultStreame
 		resultsStreamer.Done()
 	} else {
 		// otherwise execute query
-		if result, err := c.client.executeQuery(query, true, cmdconfig.Viper().Get(constants.ArgOutput) == constants.ArgTable); err != nil {
+		if result, err := c.client.executeQuery(query, cmdconfig.Viper().Get(constants.ArgOutput) == constants.ArgTable); err != nil {
```
kaidaguerre, line 208: put `cmdconfig.Viper().Get(constants.ArgOutput) == constants.ArgTable)` into a variable

kaidaguerre (changes requested):

cmd/query.go
```diff
@@ -71,6 +73,13 @@ func runQueryCmd(cmd *cobra.Command, args []string) {
 
 	queryString := getQuery(cmd, args)
 
+	// set the flag to not show spinner
+	cmdconfig.Viper().Set("show-spinner", false)
```
kaidaguerre, line 77:
```
showSpinner := queryString == ""
cmdconfig.Viper().Set("show-spinner", showSpinner)
```

cmdconfig/viper.go
```diff
@@ -28,14 +28,17 @@ func InitViper(v *viper.Viper) {
 			utils.FailOnError(err)
 		}
 	}
+
+	// set defaults
+	v.Set("show-spinner", true)
```
kaidaguerre, line 33: so we ONLY disable spinner for non interactive queries?

kaidaguerre (changes requested):

cmdconfig/viper.go
binaek89: yes.

otherwise for things like `service start` and `service refresh`, the spinner will stop working

PR #244, opened 2021-03-03
wrap `plugin list` output to window width. Closes #235

### LINKED ISSUE #235: `plugin list` should wrap table width to available window width

binaek89:


dbdefinitions/resultstreamer.go
```diff
@@ -1,30 +1,30 @@
-package db
+package dbdefinitions
```
kaidaguerre, line 1: rename to `results`

dbdefinitions/queryresults.go
```diff
@@ -1,4 +1,10 @@
-package db
+/**
```
kaidaguerre, line 1: rename file to `queryresult` (singular)

kaidaguerre (changes requested):

PR #262, opened 2021-03-04
Fix typo in install.sh

jackdelab:
There is a typo in a help message at the very end of install.sh


e-gineer:
Good catch @jackdelab - thank you!

PR #337, opened 2021-03-26
fix json encoding issues. Closes #336

### LINKED ISSUE #336: Incorrect display of string column type with value null
**Describe the bug**
the column value of string column type appears as `<null>` in the table and `"\u003cnull\u003e"` in the json output formats.

Example: v0.3.3

**To reproduce**
`steampipe query "select column_4 from chaos.chaos_get_test" --output json`

**Expected behavior**
Should be `null` instead of `"\u003cnull\u003e"`

**Additional context**
Screenshot 2021-03-26 at 3 55 39 PM

binaek89:


display/column.go
```diff
@@ -72,7 +72,7 @@ func ColumnValueAsString(val interface{}, colType *sql.ColumnType) (result strin
 // JSON, JSONB, BOOL and so on..
 func ParseJSONOutputColumnValue(val interface{}, colType *sql.ColumnType) (interface{}, error) {
 	if val == nil {
-		return constants.NullString, nil
+		return nil, nil
```
kaidaguerre, line 75: why do we remove this? We're still displaying the constants.NullString though I presume?

binaek89: If we don't remove this, then we will get something like:
```
{
  "value": "<null>"
}
```
as opposed to:
```
{
  "value": null
}
```

kaidaguerre: hmm so the null character is used for table, line and csv, but json uses actual nulls?

PR #356, opened 2021-04-07
separate `search_path` for `cli` and `service`. Closes #353

### LINKED ISSUE #353: separate `search_path` for `client` and `service`. Support search_path_prefix for client
See search path section of https://github.com/turbotio/osquery-extensions/wiki/Steampipe-mods-&-workspaces

### Setting the search path
By default, the search path puts the `public` schema first, followed by all connection schemas, ordered alphabetically.

You may change the default in the steampipe config file in the `database` options.  This will change the search path **in the database**.  This means that this path will be the default even when connecting to the Steampipe database from 3rd party tools.
```t
options "database" {
  search_path  =  "aws01,gcp01,azure01,github,slack,whois"
}
```

You may also change the default in the `terminal` options.  This will change the search path **for the session** when connecting via Steampipe, but it does NOT change the setting in the database; the search path set in `terminal` options will NOT be in effect when connecting directly to the Steampipe database from 3rd party tools. 

You may set the `terminal` options globally (in `~/.steampipe/config/default.spc`), and the setting will override the database level setting when using Steampipe.  You may also set the search path on a per-workspace basis in the `workspace.spc` file; if set, the workspace setting will override any global setting.
```t
options "terminal" {
  search_path  =  "aws01,gcp01,azure01,github,slack,whois"
}
```
 
Alternatively, you may use `search_path_prefix` in the `terminal` options.  The `search_path` is a comma-separated list of connection names that will define the **exact** search path.  `search_path_prefix` is also a comma separated list of connections, but it used to **move connections to the front** of the search path.  This is useful when you want to prefer a specific connection for one or more plugin types, but to still include the rest in the search path.  For instance, I may have many `aws` connections, and I would like to prefer `aws01` in my workspace but still leave all the other connections (`github`, `slack`, `shodan`, etc) in the search path as well.  While this is possible using `search_path`, you would need to update the `search_path` every time you install a new plugin.  This is simple and dynamic, however, when using `search_path_prefix`:

 ```t
options "terminal" {
  search_path_prefix  =  "aws01"
}
```

You may override the config file settings with the `--search-path` or `--search-path-prefix` command line arguments:
- Set the search path exactly:
```bash
steampipe query --search_path "aws_account1,gcp,azure,github"
```

- Move the `aws01` and `gcp01` connections to the front of the search path:
```bash
steampipe query --search-path-prefix "aws01,gcp01"
```

You can also set the search path for the duration of your interactive session with a the `.search_path` or `.search_path_prefix` meta-commands.  These settings will last for the duration of the session, and behave the same as the `search_path` and `search_path_prefix` options:

- Set the search path exactly:
```bash
.search_path aws_account1,gcp,azure,github
```

- move the `aws01` and `gcp01` connections to the front of the search path:
```bash
.search_path_prefix aws01,gcp01
```

You can view your current active search path by issuing the `.search_path` meta-command with no arguments
```bash
> .search_path
public, alicloud, aws, aws_01, aws_02, aws_03, azure, azure_01, digitalocean, gcp, github, hackernews, ipstack, kubernetes, oci, public, rss, slack, steampipe, whois, zendesk, internal
```

binaek89:


db/client.go
```diff
@@ -78,6 +70,41 @@ func createPostgresDbClient() (*sql.DB, error) {
...
+	var searchPath []string
+
+	if viper.IsSet("search-path") {
+		searchPath = viper.GetStringSlice("search-path")
+	} else {
+		searchPath = c.schemaMetadata.GetSchemas()
+		sort.Strings(searchPath)
+	}
+	if viper.IsSet("search-path-prefix") {
+		searchPath = append(viper.GetStringSlice("search-path-prefix"), searchPath...)
```
kaidaguerre, line 83: deduping?

could just do this
```
tmp := viper.GetStringSlice("search-path-prefix")
for _, p := range(searchPath){
  if !helpers.StringSliceContains(tmp, p){
     tmp = append(tmp, p)
}
searchPath = tmp
```

(with nicer naming....:) )

db/query.go
```diff
@@ -52,9 +52,10 @@ func ExecuteQuery(queryString string) (*results.ResultStreamer, error) {
 
 	// workaround: if connections were updated, create a new client so that any change in the search patch is reflected
 	if connectionsUpdated {
```
kaidaguerre, line 54: should no longer be needed - this is the hack I added

db/start.go
```diff
@@ -241,6 +245,46 @@ func ensureSteampipeServer() error {
 	return nil
 }
 
+func (c *Client) setServiceSearchPath() {
+	// set the search_path to the available foreign schemas or the one set by the user
```
kaidaguerre, line 249: is this comment still correct?

db/start.go
```diff
@@ -241,6 +245,46 @@ func ensureSteampipeServer() error {
...
+	// we need to do this here, since postgres resets the search_path on every load.
+	var schemas []string
+
+	if viper.IsSet("database.search-path") {
```
kaidaguerre, line 253: comment on why we use the scoped arg

db/start.go
```diff
@@ -241,6 +245,46 @@ func ensureSteampipeServer() error {
...
+		fmt.Println("got key")
+		schemas = viper.GetStringSlice("database.search-path")
+	} else {
+		schemas = c.schemaMetadata.GetSchemas()
+		// sort the schema names
+		sort.Strings(schemas)
+	}
+
+	// set this before the `public` schema gets added
```
kaidaguerre, line 262: should we set c.schemaMetadata.SearchPath when setting the service search path?

kaidaguerre: should be in setClientSearchPath surely?

kaidaguerre (changes requested):

db/start.go
binaek89: Till the `setClientSearchPath` actually gets called and sets the search path, this is the valid search path -  so setting it!

db/client.go
```diff
@@ -80,7 +82,13 @@ func (c *Client) setClientSearchPath() error {
 		sort.Strings(searchPath)
 	}
 	if viper.IsSet("search-path-prefix") {
-		searchPath = append(viper.GetStringSlice("search-path-prefix"), searchPath...)
+		prefixes := viper.GetStringSlice("search-path-prefix")
```
kaidaguerre, line 85: rename prefixes to prefixedSearchPath

db/query.go
```diff
@@ -38,7 +38,7 @@ func ExecuteQuery(queryString string) (*results.ResultStreamer, error) {
 	utils.FailOnErrorWithMessage(err, "client failed to initialize")
 
 	// refresh connections
-	connectionsUpdated, err := client.RefreshConnections()
+	_, err = client.RefreshConnections()
```
kaidaguerre, line 41: update RefreshConnections to no longer return the bool - this is the only place it was used

db/client.go
```diff
@@ -6,9 +6,11 @@ import (
 	"log"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
+	"github.com/turbot/go-kit/helpers"
+
```
kaidaguerre, line 12: remove line

db/start.go
kaidaguerre: I think it's needlessly confusing though.  The client will never _use_ this search path

db/start.go
```diff
@@ -241,6 +245,48 @@ func ensureSteampipeServer() error {
...
+		escapedSchemas = append(escapedSchemas, PgEscapeName(schema))
+	}
+
+	log.Println("[TRACE] setting search path to", schemas)
+	query := fmt.Sprintf(
+		"alter user %s set search_path to %s;",
+		constants.DatabaseUser,
+		strings.Join(escapedSchemas, ","),
+	)
+	// TODO should we report error
```
kaidaguerre, line 286: should we report error?

kaidaguerre (changes requested):

db/start.go
```diff
@@ -284,7 +280,8 @@ func (c *Client) setServiceSearchPath() {
 		strings.Join(escapedSchemas, ","),
 	)
 	// TODO should we report error
```
kaidaguerre, line 282: remove comment

db/start.go
```diff
@@ -223,9 +223,7 @@ func StartDB(port int, listen StartListenType, invoker Invoker) (StartResult, er
 		}
 	}
 
-	client.setServiceSearchPath()
-
-	return ServiceStarted, nil
+	return ServiceStarted, client.setServiceSearchPath()
```
kaidaguerre, line 226: split into 2 lines for clarity
```
err := client.setServiceSearchPath()
return ServiceStarted, err
```

kaidaguerre (changes requested):

PR #361, opened 2021-04-09
Update references from registry.steampipe.io to hub.steampipe.io

justin8:
I'm new to steampipe, but when I went to try and find plugins via the CLI it told me to go to registry.steampipe.io, which doesn't exist and doesn't redirect to the newer hub.steampipe.io; just fixing that here.


kaidaguerre:
hi @justin8. Thanks for the PR  - please could you post a screenshot of the error you are seeing.

Thanks,
kai

kaidaguerre:
Thanks for the PR  @justin8 - good catch!

PR #363, opened 2021-04-09
`search-path` and `search-path-prefix` arg in cmdline. Closes #358

### LINKED ISSUE #358: Add `search_path` and `search_path_prefix` args to query, and `search_path` arg to service start and restart
these will take precedence over search path and prefix specified in database/terminal options

see https://github.com/turbot/steampipe/issues/353

binaek89:


kaidaguerre:
replaced with https://github.com/turbot/steampipe/pull/366

PR #415, opened 2021-04-20
Ensure scheduled tasks are not executed in `query` `batch mode`. Closes #301

### LINKED ISSUE #301: Don't prompt for updates when running non-interactive query
**Describe the bug**
We should skip the update check when running queries in non-interactive mode.  If people are running scheduled jobs, this probably messes up their output...

`steampipe version 0.3.0-rc.0`

Example: v0.3.0

**To reproduce**
```sql
$ steampipe query "select * from aws_account"
+----------------------------------------------------------+
|                                                          |
| Updated versions of the following plugins are available: |
|                                                          |
|   turbot/azure        @ latest           0.2.2 → 0.3.0   |
|   turbot/digitalocean @ latest           0.2.1 → 0.3.0   |
|   turbot/alicloud     @ latest           0.0.2 → 0.0.3   |
|   turbot/aws          @ latest           0.6.0 → 0.7.0   |
|   turbot/gcp          @ latest           0.3.0 → 0.3.1   |
|                                                          |
| You can update by running steampipe plugin update --all  |
|                                                          |
+----------------------------------------------------------+

```
**Expected behavior**
No update prompts for non-interactive query - just the results.

binaek89:


task/runner.go
```diff
@@ -22,21 +24,22 @@ func RunTasks() {
 func NewRunner() *Runner {
 	r := new(Runner)
 	r.currentState, _ = statefile.LoadState()
-	r.shouldRun = r.getShouldRun()
 	return r
 }
 
 func (r *Runner) Run() {
-	if r.shouldRun {
+	if r.getShouldRun() {
```
kaidaguerre, line 31: rename to `shouldRun()`

task/runner.go
```diff
@@ -66,3 +69,14 @@ func (r *Runner) getShouldRun() bool {
 	minutesElapsed := now.Sub(lastCheckedAt).Minutes()
 	return minutesElapsed > minimumMinutesBetweenChecks
 }
+
+func (r *Runner) shouldRunUpdateChecks() bool {
```
kaidaguerre, line 73: add a comment - return whether to run the update checks. These are disabled when running queries in batch mode.

add a similar comment for shouldRun

kaidaguerre (changes requested):

PR #419, opened 2021-04-23
New Feature :: `CI` mode for steampipe commands. Closes #233

### LINKED ISSUE #233: steampipe `CI` mode
We should have a `--ci` flag for all commands.

`TBD: do we also need an environment variable?`

When `CI` mode is enabled, the CLI will:
* disable `scheduled tasks` (`cli update check`, `plugin update check`, `log trimming`)
* disable `spinner` display
* disable `paging` of output

binaek89:
`CI` mode in the steampipe CLI is meant to be useful when the user wants to disable all extraneous activities.

We have taken design queues from `terraform` which has `-input=false` and `-auto-approve` flags for running in automated environments. `terraform` also support a `TF_IN_AUTOMATION` environment variable.

This PR adds a new `--ci` flag for all commands.

When the `--ci` flag is passed in any command, the CLI will:
* disable `scheduled tasks` (`cli update check`, `plugin update check`, `log trimming`)
* disable `spinner` display
* disable `paging` of output

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -194,7 +194,7 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	// a leading blank line - since we always output multiple lines
 	fmt.Println()
 
-	spinner := utils.ShowSpinner("")
+	spinner := display.ShowSpinner("")
 
 	for _, p := range plugins {
 		isPluginExists, _ := plugin.Exists(p)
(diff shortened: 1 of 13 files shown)


e-gineer:
@binaek ... This is a fairly large change to the command line behavior, will require significant review of the commands level (not so much the code).

Can you please expand on exactly how this new command will work? How does it interact with / change the existing commands? Looking to understand the changes to the commands / behavior / user experience.

PR #429, opened 2021-04-26
Ensure update notification is never displayed when running in query `batch` mode. Closes #428

### LINKED ISSUE #428: `update-check` notification in `batch` `query` mode
**Describe the bug**
Update checks take place even if in `batch mode` when the `internal/update-checks.json` file does not exist

v0.4.1

**To reproduce**
* move existing `internal/update-checks.json` to a known location
* execute `steampipe query "select 1 val"`

In the output, update notification is coming up before the query output

**Expected behavior**
Should print only the query output:
```
+-----+
| val |
+-----+
| 1   |
+-----+
```

**Additional context**

binaek89:


task/runner.go
```diff
@@ -58,26 +58,33 @@ func (r *Runner) runAsyncJob(job func(), wg *sync.WaitGroup) {
 // determines whether the task runner should run at all
 // tasks are to be run at most once every 24 hours
 // also, this is not to run in batch query mode
-func (r *Runner) shouldRun() bool {
+func (r *Runner) shouldRun() (returnValue bool) {
+	returnValue = true
+
+	defer func() {
```
kaidaguerre, line 64: why defer this - why not just return out at once in batch mode?

also - not too keen on usage of named return value here - I think often it makes code more confusing. I guess we need it for defer... but could be removed if we remove defer?

PR #432, opened 2021-04-27
.output metacommand should more helpful output when no args are passed. closes #375

### LINKED ISSUE #375: .output metacommand should more helpful output when no args are passed
currently it shows 
```
Error: command needs 1 argument(s) - got 0
```

It should show something useful like `.multi` does
```
multi-line mode is on. You can disable it with: .multi off
```

tyagiparth:


kaidaguerre:
A more extensive approach to this is required, closing this for now until we have time to revisit

PR #458, opened 2021-05-11
fixes spinner new line issue for long texts. Closes #450

### LINKED ISSUE #450: Spinner displays multiple newlines if spinner text is wider than the terminal
for example 
```
steampipe check aws.control_group.cis_v130
```
with a narrow screen

binaek89:
spinner displays a new line on every refresh if the text given as suffix to the spinner exceeds the current terminal width.

this fix truncates the suffix to the maximum available width and appends ellipsis at the end

### DIFF
--- a/display/spinner.go
+++ b/display/spinner.go
@@ -1,4 +1,4 @@
-package utils
+package display
 
 import (
 	"fmt"
--- a/display/spinner.go
+++ b/display/spinner.go
@@ -3,6 +3,7 @@ package display
 import (
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/briandowns/spinner"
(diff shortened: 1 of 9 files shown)


display/spinner.go
```diff
@@ -10,16 +10,27 @@ import (
...
 
+//
+// spinner format:
+// <spinner><space><message><space><dot><dot><dot><cursor>
+// 		1	   1   [.......]   1     1    1    1     1
+// We need at least seven characters to show the spinner properly
+//
+// Not using the (…) character, since it is too small
+//
+const mustHaveCharLengthForSpinner = 7
```
kaidaguerre, line 21: rename to minSpinnerWidth

display/spinner.go
```diff
@@ -10,16 +10,27 @@ import (
...
+
 func truncateSpinnerMessageToScreen(msg string) string {
 	if len(strings.TrimSpace(msg)) == 0 {
 		// if this is a blank message, return it as is
 		return msg
 	}
+
 	maxCols, _, _ := gows.GetWinSize()
-	if maxCols < 7 {
+	if maxCols < mustHaveCharLengthForSpinner {
```
kaidaguerre, line 30: so is this check saying - 
```
if the screen is smaller than the minimum spinner width, we cannot truncate to make it fit
```

if so, let's add that as a comment

PR #476, opened 2021-05-18
Add support for control dry run with a `--dry-run` flag to check command to show which controls would be run. Closes #468

### LINKED ISSUE #468: Add --dry-run flag to check command to show which controls would be run

binaek89:
Screenshot


kaidaguerre (changes requested):
think a simple list of controls to be run would be better - lets lose the dots and the result could and square brackets etc

kaidaguerre:
from John: 
```
yes, I think the hierarachy with just the titles (without the dots and totals which are always 0), indented is probably a good starting place

```

PR #489, opened 2021-05-20
Add support for inspection and autocomplete of Reflection tables. Closes #446

### LINKED ISSUE #446: Implement inspect for reflection tables

binaek89:
`interactive` autocomplete has reflection tables as `unqualified` tables
Screenshot 2021-05-20 at 6 01 02 PM

`.inspect <reflection_table_name>` shows the relevant table:
Screenshot 2021-05-20 at 6 01 56 PM

`.inspect` without arguments does not show the `temp` schema:
Screenshot 2021-05-20 at 6 03 29 PM

### DIFF
--- a/autocomplete/table.go
+++ b/autocomplete/table.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/c-bata/go-prompt"
 	"github.com/turbot/go-kit/helpers"
-	"github.com/turbot/steampipe/constants"
 	"github.com/turbot/steampipe/schema"
 	"github.com/turbot/steampipe/steampipeconfig"
 )
@@ -20,7 +19,7 @@ func GetTableAutoCompleteSuggestions(schema *schema.Metadata, connectionMap *ste
 	// schema names
 	var schemasToAdd []string
 	// unqualified table names - initialise to the reflection table names
-	unqualifiedTablesToAdd := constants.ReflectionTableNames()
+	unqualifiedTablesToAdd := []string{} // constants.ReflectionTableNames()
 	// fully qualified table names
 	var qualifiedTablesToAdd []string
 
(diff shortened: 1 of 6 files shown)


db/schema.go
```diff
@@ -50,9 +49,13 @@ func buildSchemaMetadata(rows *sql.Rows) (*schema.Metadata, error) {
...
 			Description: record.ColumnDescription,
 		}
+
+		if strings.HasPrefix(record.TableSchema, "pg_temp") {
+			schemaMetadata.TemporarySchemaName = record.TableSchema
+		}
 	}
 
-	return schemaMetadata, err
+	return schemaMetadata.Schemas, schemaMetadata.TemporarySchemaName, err
```
kaidaguerre, line 58: why can't we just return schemaMetadata as before

query/metaquery/handlers.go
```diff
@@ -251,6 +254,9 @@ func inspect(input *HandlerInput) error {
 		if !schemaFound {
 			searchPath := input.Schema.SearchPath
 
+			// add the temporary schema to the search_path so that it becomes searchable
+			searchPath = append(searchPath, input.Schema.TemporarySchemaName)
```
kaidaguerre, line 258: why do we need this?

autocomplete/table.go
```diff
@@ -20,7 +19,7 @@ func GetTableAutoCompleteSuggestions(schema *schema.Metadata, connectionMap *ste
 	// schema names
 	var schemasToAdd []string
 	// unqualified table names - initialise to the reflection table names
-	unqualifiedTablesToAdd := constants.ReflectionTableNames()
+	unqualifiedTablesToAdd := []string{} // constants.ReflectionTableNames()
```
kaidaguerre, line 22: remove commented out code. Do we still need constants.ReflectionTableNames()?

kaidaguerre (changes requested):

autocomplete/table.go
binaek89: We still need it. There are other places which refer to the `constants.ReflectionTableNames()` function.

query/metaquery/handlers.go
binaek89: The iteration happens through the current search path - but the temporary schema is not a part of that list. Need to add it to the end.

PR #515, opened 2021-05-28
Add support for export flag to check command. Closes #511

### LINKED ISSUE #511: Add support for `export` flag to check command
### Export Output Formats

- Export formats are fixed - there are no options other than the file name.

- There is no 'text' format - redirect or `tee` stdout if you want the text. 

- Exports don't usually include `documentation` or `sql` in the output,  These fields are very large and of questionable value here...

- Usually, exports should to use 'full' name (`aws.benchmark.cis_v130`), since you could run benchmarks from multiple mods.

| Format | Description
|-|-
| `csv` | Comma separated, one line per control.  Contains all the control metadata, "flattened" and repeated for queries that return multiple rows.  No summary data. Order not guaranteed.
| `json` | json data.  Contains all control meta-data, detail, and summary data
| `html` | single-file html report.  Uses javascript to show/hide items, but contains all control meta-data, detail, and summary data, in order ( by default)
| `junit-xml` | XML in junit format, which includes nesting and totals / sub-totals
| `pdf` | full report, with control meta-data, detail, and summary data, in order
| `table` | raw results, inserted into a new table in the public schema.  Contains all control meta-data and detail.  No summary data.  This is essentially the CSV format, imported into a table

- which are in scope: 
  - json
  - csv

- Specify outputs via one or more `--export` statements.  
  - format is '{format name}[:{file path}]
    - {format name} is a specific named format (`csv`, `json`, etc)
    - A user can specify the path to export to by including it after a colon.
    - if no path is specified, the file will be written to the current working directory, with a name that includes a timestamp (to prevent accidentally over-writing files).

```bash
steampipe check all --export=csv --export=json:myfile.json --export=html:./index.html --export=table:todays_fresh_baked_controls

```

```bash
# default interactive console out, no export to files
steampipe check aws

### PR DESCRIPTION
**Pull request by binaek89**
This needs to be merged to `main` after merging #506

### CODE REVIEW

File: cmd/check.go [hunk 1]
```diff
@@ -105,7 +109,17 @@ func runCheckCmd(cmd *cobra.Command, args []string) {
 
 	// treat each arg as a separate execution
 	failures := 0
+	exportWaitGroup := sync.WaitGroup{}
 	for _, arg := range args {
+
+		// get the export formats for this argument
+		exportFormats := getExportFormats(arg)
+		err = validateExportFormats(exportFormats)
+		if err != nil {
```
kaidaguerre, line 118: can we just use FailOnError

cmd/check.go
```diff
@@ -152,58 +189,63 @@ func initialiseColorScheme() error {
...
 
-func displayTextOutput(ctx context.Context, executionTree *execute.ExecutionTree) error {
-	maxCols := getMaxCols()
-
-	renderer := controldisplay.NewTableRenderer(executionTree, maxCols)
-
-	if ctx.Err() != nil {
-		utils.ShowError(ctx.Err())
+func exportControlResults(ctx context.Context, executionTree *execute.ExecutionTree, formats []controldisplay.CheckExportFormat) error {
+	fmt.Println()
```
kaidaguerre, line 204: if output format is none, we should not print anything? Let's verify with John

cmd/check.go
```diff
@@ -152,58 +189,63 @@ func initialiseColorScheme() error {
...
+				fileName = parts[1]
+			}
+		}
+		formats = append(formats, controldisplay.NewCheckOutputFormat(format, fileName))
 	}
-	return maxCols
+	return formats
+}
+
+func generateDefaultExportFileName(format string, executing string) string {
```
kaidaguerre, line 249: does John specify a file format and does this match it - or have you come up with this yourself?

kaidaguerre (changes requested):

kaidaguerre (changes requested):
Please add BATS test

cmd/check.go
```diff
@@ -157,6 +181,16 @@ func validateOutputFormat() error {
 	return nil
 }
 
+func validateExportFormats(formats []controldisplay.CheckExportFormat) error {
+	for _, exportFormat := range formats {
+		_, err := controldisplay.GetExportFormatter(exportFormat.Format)
```
kaidaguerre, line 186:
```
if _, err := controldisplay.GetExportFormatter(exportFormat.Format); err != nil {
			return err
		}
```

kaidaguerre (changes requested):

PR #516, opened 2021-05-31
Add support for reading from `STDIN` for `query`. Closes #499

### LINKED ISSUE #499: steampipe accepts the query on `STDIN`
Today, when running a query from the command line, `steampipe query` takes the query as parameter. When composing pipelines, I would like to provide the query on `STDIN`:

```command
cat <<EOQ | steampipe query --output json                                                                                                                                           
  select
    title, url
  from
    hackernews_top
  order by
    score desc
  limit
    10
EOQ
```

This is already possible with `psql` (after `steampipe service start`):

```command
cat <<EOQ | psql 'postgres://steampipe:********@localhost:9193/steampipe?sslmode=disable'
  select
    title, url
  from
    hackernews_top
  order by
    score desc
  limit
    10
EOQ
```

In the examples above, `cat` is used as a placeholder for other commands generating a query.

Thanks for `steampipe`, please keep up the good work!

binaek89:


kaidaguerre (commented):
I presume you've tested this, piping input on a bash command line?

kaidaguerre (changes requested):
I presume you've tested this, piping input on a bash command line?

Please add BATS test

PR #518, opened 2021-05-31
Fix to display timing profile data when appropriate env var is set.  Closes #513

### LINKED ISSUE #513: get time logging working again

binaek89:


main.go
```diff
@@ -7,30 +7,31 @@ import (
...
 	"github.com/turbot/steampipe/cmd"
 	"github.com/turbot/steampipe/utils"
 )
 
 var Logger hclog.Logger
 
 func main() {
-	logging.LogTime("main start")
+	utils.LogTime("main start")
+	exitCode := 0
```
kaidaguerre, line 18: there is a global exit code in root, so use that.

And this code in root: 

```
func Execute() {
	rootCmd.Execute()
	os.Exit(exitCode)
}
```
maybe that just needs to be a defer

kaidaguerre (changes requested):

main.go
binaek89: If we exit from `root.go` then the `defer` in `main.go` doesn't fire and the `DisplayProfileData` function is never executed.

PR #547, opened 2021-06-16
Support for infering export format when only the target file is given. Closes #543

### LINKED ISSUE #543: infer the export file type by the suffix
**Describe the bug**
we should infer the export file type by the suffix.  This is my fault - the original doc says to use the 'type:fileename' syntax which is what is implemented, but it was superseded by the 'controls ref' doc and i failed to note that .  Hopefully this is a minor change...
  - was implemented as: `steampipe check benchmark.cis_v130_2_1 --export json:x.json --export csv:x.csv`  
  - should instead be : `steampipe check benchmark.cis_v130_2_1 --export x.json --export x.csv`
  
main, commit e5b99605cf49dca140d2d83227bc7e59690193b7

binaek89:


cmd/check.go
```diff
@@ -243,14 +243,35 @@ func getExportFormats(executing string) []controldisplay.CheckExportFormat {
 	formats := []controldisplay.CheckExportFormat{}
 	exports := viper.GetStringSlice(constants.ArgExport)
 	for _, export := range exports {
+		if len(strings.TrimSpace(export)) == 0 {
+			// if this is an expty string, ignore
```
kaidaguerre, line 247: typo

cmd/check.go
```diff
@@ -243,14 +243,35 @@ func getExportFormats(executing string) []controldisplay.CheckExportFormat {
...
+			format = parts[0]
+			fileName = parts[1]
+			if f, err := helpers.Tildefy(fileName); err == nil {
 				fileName = f
-			} else {
-				fileName = parts[1]
+			}
+		} else {
+			format = parts[0]
+			fileName = generateDefaultExportFileName(format, executing)
```
kaidaguerre, line 265: I find this flow a little confusing - I think this is clearer

```
                         format = parts[0]

			// try to get an export formatter
			if _, err := controldisplay.GetExportFormatter(format); err  != nil {
				// this is not a valid format. assume it is a file name
				fileName = format

				// now infer the format from the file name
				format = controldisplay.InferFormatFromExportFileName(fileName)
			} else {
                                 // the format is valid, generate a filename
                                 fileName = generateDefaultExportFileName(format, executing)
                          }
```

control/controldisplay/formatter.go
```diff
@@ -62,6 +63,16 @@ func GetOutputFormatter(outputFormat string) (Formatter, error) {
 	return formatter, nil
 }
 
+func InferFormatFromExportFileName(filename string) string {
+	extension := strings.TrimPrefix(filepath.Ext(filename), ".")
+	switch extension {
+	case "csv", "json":
+		return extension
+	default:
+		return ""
```
kaidaguerre, line 72: if we fail to infer a format, that should be an error right?

kaidaguerre (changes requested):

cmd/check.go
```diff
@@ -243,14 +243,35 @@ func getExportFormats(executing string) []controldisplay.CheckExportFormat {
...
-			if f, err := helpers.Tildefy(parts[1]); err == nil {
+
+		format := ""
+		fileName := ""
+
+		if len(parts) == 2 {
+			// we have two distinct parts - life is good
+			format = parts[0]
+			fileName = parts[1]
+			if f, err := helpers.Tildefy(fileName); err == nil {
```
kaidaguerre, line 260: what happens in case of error?

binaek89: We keep it as it is and try to work with it.

If the `fileName` is not resolvable, `os` will error out when we try to create it

kaidaguerre: yeah I don't like that really - got a screen shot of how that looks?

I think I'd rather handle it explicitly

If it is to remain implicit, we need a comment

PR #564, opened 2021-06-17
Do not show spinner when updating metadata tables after a file watcher event. Closes #563

### LINKED ISSUE #563: Do not show spinner when updating metadata tables after a file watcher event

kaidaguerre:
…

### DIFF
--- a/control/execute/control_run.go
+++ b/control/execute/control_run.go
@@ -96,7 +96,7 @@ func (r *ControlRun) Start(ctx context.Context, client *db.Client) {
 	}
 
 	// set a log line in the database logs for convenience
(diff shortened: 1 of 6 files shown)


control/execute/control_run.go
```diff
@@ -96,7 +96,7 @@ func (r *ControlRun) Start(ctx context.Context, client *db.Client) {
 	}
 
 	// set a log line in the database logs for convenience
-	_, _ = client.ExecuteSync(ctx, fmt.Sprintf("--- Executing %s", *control.Title))
+	_, _ = client.ExecuteSync(ctx, fmt.Sprintf("--- Executing %s", *control.Title), false)
```
binaek89, line 99: This can be set to `true`. This is used to set a comment in the database log for convernience

PR #601, opened 2021-07-12
`plugin install` creates `temp` files in `$STEAMPIPE_INSTALL_DIR/plugins` during plugin install. Closes #600. Closes #527

### LINKED ISSUE #527: plugin install fails when .steampipe/tmp is on a different partition than .steampipe/plugins
**Describe the bug**
plugin install fails when .steampipe/tmp is on a different partition than .steampipe/plugins. 
0.5.1

**To reproduce**
mount ~/.steampipe/tmp on a separate partition than ~/.steampipe/tmp and attempt to install a plugin

**Expected behavior**
The plugin should be installed.  

**Additional context**
 I had thought this was fixed previously, but I couldnt find the issue.  From the code, it looks like its still doing a 'rename' style move instead of a copy and delete.  For example:  https://github.com/turbot/steampipe/blob/main/ociinstaller/plugin.go#L100

### LINKED ISSUE #600: `plugin install` should create it's `temp` directory in `plugin` directory

binaek89:


ociinstaller/tempdir.go
```diff
@@ -21,7 +23,7 @@ func NewTempDir(path string) *tempDir {
 }
 
 func getOrCreateTempDir(ref string) string {
-	pluginCacheDir := filepath.Join(constants.TempDir(), safeDirName(ref))
+	pluginCacheDir := filepath.Join(constants.PluginDir(), safeDirName(fmt.Sprintf("tmp-%s", generateTempDirName())))
```
kaidaguerre, line 26: do we delete this dire afterwards?

binaek89: Yes.

In `ociinstaller/plugin.go` `#L18`

PR #609, opened 2021-07-13
`initdb` even if databasse is installed but `data directory` is empty. Closes #575

### LINKED ISSUE #575: run initdb if the postgres data dir is empty
When running from docker, you may want to pre-install steampipe binaries and plugins in the container image and run them read only, but the data directory must be read write.  Currently, if you mount that dir as a docker bind mount, or a kube volume, steampipe dant start because the data dir is empty.

run initdb to initialize and empty db in the data dir if the postgres data dir is empty.

binaek89:


cmd/service.go
```diff
@@ -435,53 +435,66 @@ To keep the service running after the %s session completes, use %s.
...
-
 	force := cmdconfig.Viper().GetBool(constants.ArgForce)
 	if force {
-		status, err = db.StopDB(force, db.InvokerService)
+		status, err = db.StopDB(force, db.InvokerService, spinner)
 	} else {
-		info, err := db.GetStatus()
+		info, err = db.GetStatus()
 		if err != nil {
+			display.StopSpinner(spinner)
```
kaidaguerre, line 459: can we not do this in the defer - check if the spinner is nil and if not stop it

kaidaguerre (changes requested):

cmd/service.go
binaek89: The spinner has to be stopped before printing anything else.

If we do the stop in `defer` then we will have to float up the message as well

PR #640, opened 2021-07-16
Fixes issue where `data-dir` was being removed instead of it's contents during `init`. Closes #639

### LINKED ISSUE #639: `init` should remove `data-dir` contents and not the `data-dir`

binaek89:


utils/remove_dir_contents.go
```diff
@@ -0,0 +1,20 @@
+package utils
+
+import (
+	"os"
+	"path/filepath"
+)
+
+func RemoveDirectoryContents(removePath string) error {
```
kaidaguerre, line 8: if removePath does not exist, what happens?

PR #767, opened 2021-08-10
Remove hardcoded runtime flags by adding postgresql.conf.d folder

tylarb:
Steampipe hard coded some performance related runtime flags in the
server start configuration. This means they aren't editable after the
binary is compiled.

This commit appends the all flags which were previously not modifyable
to a steampipe.conf file, and adds a postgresql.conf file, which loads
the steampipe.conf and all files in the postgresql.conf.d folder.
    
Custom settings or overrides of the default values can be set in any
file ending in ".conf" in the postgresql.conf.d folder if the user
chooses to do so afterwords.

The flags which are editable in a steampipe config, such as SSL and log
location (as an extension of data location) remain.

Signed-off-by: Tyler Ramer <tramer@yugabyte.com>


e-gineer:
Hey @tylarb ... thanks for your contribution here! Agree it will be helpful to add more flexibility around the postgres settings and this looks like a good approach for doing that 👍 

Couple of questions:

Do you have a specific case for making these options changeable? I'm wondering about the motivation / use case ... mostly to help us better understand how to make Steampipe as useful as possible :-)

Did you happen to do any review of the actual settings? We'd love feedback on whether these are appropriate settings for running an embedded postgres instance. TBH, they seemed like a good starting point from our reading of various sources but we'd love any suggestions!

tylarb:
Use case:
We've connected steampipe to Zendesk, and since we've got the API (and thus tickets) available as data, we create an enriched table that then adds further data (things like Jira information, supplementary data). This enriched table is write heavy and of course we query several other tables to created the enriched data, and we do a sort of ETL (dump of the Zendesk tickets) every 1 hour - and these default settings which you've enabled are somewhere between poor and really bad for write / insert and general "big data" workloads

Review of settings:
I'm coming from the big data side of the house, and have plenty of hardware  to throw at this problem (relative to our needs anyway - it's just a 4 core cloud VM / 8 GB ram or similar) and value query response time (including inserts) and data resiliency more than minimalism. So I don't really have much to say about the settings except that they're probably fine for your intended use case - full admission we may be overloading your product some. 

Only ones I'd note for sure - and the ones I paid attention to - [turning off fsync risks data corruption](https://www.2ndquadrant.com/en/blog/postgresql-fsync-off-warning-in-config-file/) and you set most of these to basically be really poor at write performance and avoid bloat - turning off autovacuum, tuning down work_mem and wal_buffers. That's ok - I just wanted to be able to change them myself, not change your product direction or vision. 

I discussed with my teammate, and thought it could be good to separate steampipe from shipping a postgres, and using an off-the-shelf postgres instance if one was available, using only the foreign data wrapper and supplying our own postgres (or Yugabyte, as it may be). But in the meantime, we'll use the shipped / downloaded postgres and at least make it sane for moderately write / query intensive workloads.

e-gineer:
An exciting use case to hear about - thank you! Your description makes perfect sense ... and I agree that our out of the box settings are definitely not optimized for write performance, or goal was to make it small and light for desktop FDW use. I suspect there is more we can do to get the balance right, and no doubt this PR will make that much easier to tune.

Thanks again - we'll take a look at getting this merged 👍

binaek89:
Hello @tylarb 

Thank you for your contribution. As @e-gineer mentioned, the underlying `postgres` in `steampipe` was never really optimised for write performance. We went for lean-and-fast FDW reads at it's expense - leaving a lot to be desired.

The PR looks great and is definitely a step forward. I do have one concern though.

The change will (probably) break current installations, since the `initDatabase` in `db/local_db/install.go` will not get triggered in systems where `steampipe postgres` is already installed - hence the `postgresql.conf` file will not get written and `startPostgresProcess` in `db/local_db/start_database.go` does not have the necessary `-c flags` anymore.

I think if you move the changes in `db/local_db/install.go` to `db/local_db/start_database.go#L113` right before `startPostgresProcess` is called, it will migrate properly, without breaking current installations.

tylarb:
@binaek89  

This is somewhat a challenging problem. 

Existing installations won't break - they'll still run, just sans the recommended flags, using default values. The `postgresql.conf` file is generated on `initdb`, so existing installations have an existing `postgresql.conf` file that's blank (it's the "default" values that are all commented out before my additions on line 754 - https://github.com/turbot/steampipe/pull/767/files#diff-8b453dbbe24b15d5e667b3f91b720cf14c22451c1f3a6ea3363ee4ae5fce2350R754  )


Unfortunately, the command line flags override anything in `postgresql.conf` - which is why I can't just use the existing configuration and modify the `postgresql.conf` that was generated on `initdb`. 

https://www.postgresql.org/docs/9.5/config-setting.html
> Settings provided in this way override those set via postgresql.conf or ALTER SYSTEM, so they cannot be changed globally without restarting the server.

It won't work to add these changes `startPostgresProcess` because in that case, the `postgresql.conf` would be overwritten every time the DB is restarted (a `steampipe service restart`), which is the exact problem  that exists now, except that run-time changable flags would at least be modifyable with an `SET` command in the database, but some of the current flags - like `fsync` can only be set on server start, and aren't runtime editable - https://www.postgresql.org/docs/8.1/runtime-config-wal.html 

Also, if we somehow edited and applied a "clean slate" `postgresql.conf` for one upgrade only, you run the risk (perhaps small) of breaking  / editing the config for users who have set some custom values in `postgresql.conf` already


That said, removing these flags isn't really bad. Everything just goes back to the default for existing installation, but nothing should really break. It's just maybe not optimal for tiny machines that don't ever do inserts / writes.

e-gineer:
Agree it's important that we don't overwrite the `postgresql.conf` file after initial install (and potential modification). Seems like we may need to find a way to detect if it's already installed and only set it (to the default) if not already there.

tylarb:
> to detect if it's already installed 

Again, anytime initdb is run, the `postgresql.conf` will exist - so if you can start the DB, the file exists. So in every installation, the file exists. My PR just immediately overwrites it because it's still inside the initdb section, the same way you're over-riding `pg_hba.conf` - which is also created with default values on initdb completion.  

I think that modification would have to be part of out-of-band upgrade - do you have any code related to that currently? (normally, that would be handled by a package manager like RPM or DEB)

e-gineer:
Yes @tylarb, you are exactly right. We have the `postgresql.conf` file anyway (even today), so the real question is trying to work out how to avoid fighting with custom configuration settings a user may want.

Perhaps we can find a good model through `include`, `include_if_exists` or `include_dir` per https://www.postgresql.org/docs/12/config-setting.html#CONFIG-INCLUDES

Perhaps we could:
* Add a single `include_dir` line to the end of `postgresql.conf`
* Place steampipe specific settings into a `<confdir>/steampipe.conf` file
* Document this approach, encouraging users to add their own `*.conf` files before or after `steampipe.conf` in the `<confdir>`.

This makes it easier to manage our own settings (Steampipe owns the `steampipe.conf` file) and allows users to manage their own files in the setup.

Other areas we need to think through here:
* Documentation of the setup - make sure it's easily understood how to safely set these options.
* Support for docker, particularly the directory location of the postgresql.conf file and `<confdir>`.

tylarb:
@e-gineer  I think the idea of a include_dir for the settings is good. It's a clear demarcation for the settings so that's good. I'm not seeing a way that it solves the upgrade problem.

Even if steampipe forces it's steampipe.conf management (in the startup code for example, instead of initdb) how do we get the include_dir into the postgresql.conf file ?

Yes, I think last setting wins but it's not as explicit as "one setting". 


Does documentation need to be part of this pr ? (Is documentation in this repo?) 

Also - can you explain more on the docket consideration?

e-gineer:
I think we can just take the risk to overwrite the `postgresql.conf` to add the `include_dir`. We've never officially supported editing it before now, and I suspect it's very rare. If we do that with a clear direction on how to setup conf files going forward I believe that will work well. (But you are likely to know better than me on this @tylarb <grin>).

The documentation is not currently open source (mostly due to time / convenience). We can internally update this as part of the release, but any drafts in this issue etc would of course be a great helpful starting point.

For docker, we have [using containers docs](https://steampipe.io/docs/develop/containers). The main question is the location of the postgresql.conf file and ensuring it's mapped to a local file system. I don't think this will be a major item, just a note that it's part of the `./db/12.1.0/data/postgresql.conf` directory.

tylarb:
@e-gineer  I'll work on a changed patch here. I have some ideas to make it work pretty cleanly given your suggestions - I think I'd be ok with the current patch (and the pre-update breaks it may cause) but understand not wanting to do that, and I think I can hack together a solution that works.

binaek89:
@tylarb we want to put this into our next release.

Have you been able to get back to this?

Do you have an update that we can run with? If not, we are happy to pick it up from your PR and make the necessary changes.

tylarb:
I pushed a fixup (and another fixup...) with an approach to keeping both static runtime flags and directions to add custom / override configurations in the postgresql.conf.d folder (which is also created).

Notes... os.MkdirAll _already_ [doesn't fail if there's an existing dir](https://pkg.go.dev/os#MkdirAll), so I used that without the  extra checks you have in the `locations.go` package where you stat the dir first, etc.

Let me know what you think, and if it's good, I'll squash, ammend the initial comment so that it's more correct.

There should be a rough "documentation" in place in the config files as well at least "don't edit me," and "do edit this"

e-gineer:
Thanks so much @tylarb for the quick turnaround! We'll be working to get this included into the next release 👍

tylarb:
@e-gineer  if it looks good enough to modify the commit message let me know. The structure has changed since the initial message I pushed.

kaidaguerre (approved):

kaidaguerre:
@tylarb many thanks for this, it looks great, very clean and consise.

Please go ahead and squash/clean up, and we can go ahead and get this merged.

tylarb:
Done

PR #772, opened 2021-08-12
Add html formatter. closes #480

### LINKED ISSUE #480: html output format

pskrbasu:
WIP


control/controldisplay/formatter.go
```diff
@@ -52,7 +54,7 @@ type Formatter interface {
 func GetExportFormatter(exportFormat string) (Formatter, error) {
 	formatter, found := exportFormatters[exportFormat]
 	if !found {
-		return nil, fmt.Errorf("invalid export format '%s' - must be one of json,csv", exportFormat)
+		return nil, fmt.Errorf("invalid export format '%s' - must be one of json,csv,html", exportFormat)
```
kaidaguerre, line 57: can we put valid formatters into a consta string array, then we can use it here and in `InferFormatFromExportFileName`

control/controldisplay/formatter_html.go
```diff
@@ -0,0 +1,32 @@
...
+//go:embed html_template/*
+var templateFS embed.FS
+
+func (j *HTMLFormatter) Format(ctx context.Context, tree *controlexecute.ExecutionTree) (io.Reader, error) {
+	t, err := template.ParseFS(templateFS, "html_template/*")
+	if err != nil {
+		return nil, err
+	}
+	reader, writer := io.Pipe()
+	go func() {
```
kaidaguerre, line 23: update to:

```
go func() {
		if  err := t.Execute(writer, tree); err != nil {
			writer.CloseWithError(err)
		} else {
		        writer.Close()
                 }
	}()
```

kaidaguerre (changes requested):

PR #826, opened 2021-08-31
Small updates to `steampipe service *` commands. Closes #812

### LINKED ISSUE #812: Small suggested updates to the `steampipe service *` output
For steampipe service here is the sequence of output I see:
```
~/src/steampipe-plugin-fastly $ steampipe service start

Steampipe database service is now running:

	Host(s):  localhost, 127.0.0.1, 192.168.7.220
	Port:     9193
	Database: steampipe
	User:     steampipe
	Password: 0068-XXXX-XXXX
	SSL:      on

Connection string:

	postgres://steampipe:0068-XXXX-XXXX@localhost:9193/steampipe?sslmode=require

Managing Steampipe service:

	# Get status of the service
	steampipe service status
	
	# Restart the service
	steampipe service restart

	# Stop the service
	steampipe service stop
	

~/src/steampipe-plugin-fastly $ 
~/src/steampipe-plugin-fastly $ steampipe service stop
Steampipe database service stopped
~/src/steampipe-plugin-fastly $ steampipe service status
Steampipe database service is NOT running
~/src/steampipe-plugin-fastly $ 
```

Suggestions:
* Use two spaces instead of tabs to indent the text.
* Remove `?sslmode=require`, it's preferred anyway and we don't technically require it.
* Remove `SSL: on` line, we provide no way to disable it anyway.
* Double line gap at the end should be a single line gap only.
* `Managing Steampipe service:` should be `Managing the Steampipe service:`
* References to `database service` should just be `service` (simpler)
* `Steampipe database service stopped` message should be `Steampipe service stopped [port 9193].`
* `Steampipe database service is NOT running` message should be `Steampipe service is not running [port 9193].`

binaek89:


cmd/check.go
```diff
@@ -100,6 +101,7 @@ func runCheckCmd(cmd *cobra.Command, args []string) {
 		utils.LogTime("runCheckCmd end")
 		if r := recover(); r != nil {
 			utils.ShowError(helpers.ToError(r))
+			debug.PrintStack()
```
kaidaguerre, line 104: do we want to print the stack here?

db/db_local/running_info.go
```diff
@@ -27,6 +28,17 @@ func (r *RunningDBInstanceInfo) Save() error {
 	return saveRunningInstanceInfo(r)
 }
 
+func (r *RunningDBInstanceInfo) String() string {
+	writeBuffer := bytes.NewBufferString("")
+	jsonEncoder := json.NewEncoder(writeBuffer)
+	p := r.Password
```
kaidaguerre, line 34: add comment to explain what we are doing here

kaidaguerre (changes requested):

db/db_local/running_info.go
binaek89: added

cmd/check.go
binaek89: no we don't need this. removed

cmd/service.go
```diff
@@ -491,32 +495,30 @@ func printStatus(info *db_local.RunningDBInstanceInfo) {
...
+  steampipe service restart
 
-	# Stop the service
-	steampipe service stop
-	
+  # Stop the service
+  steampipe service stop
 `
-		statusMessage = fmt.Sprintf(msg, strings.Join(info.Listen, ", "), info.Port, info.Database, info.User, info.Password, db_local.SslStatus(), info.User, info.Password, info.Listen[0], info.Port, info.Database, db_local.SslMode())
+		statusMessage = fmt.Sprintf(msg, strings.Join(info.Listen, ", "), info.Port, info.Database, info.User, info.Password, info.User, info.Password, info.Listen[0], info.Port, info.Database)
```
kaidaguerre, line 521: still includes password?

PR #891, opened 2021-09-13
Fixes service lifecycle management issues when state file is deleted which service is running. Closes #872

### LINKED ISSUE #872: `service` cannot be used if `state-file` is deleted while `service` is running
While the service is running, if the state file in `STEAMPIPE_INSTALL_DIR/internal/steampipe.json` is deleted, the service cannot be connected to anymore.

Additionally, `service stop --force` does not work in this state as well.

binaek89:


db/local_db/stop_database.go
```diff
@@ -22,7 +22,7 @@ type StopStatus int
 
 const (
 	// ServiceStopped :: StopStatus - service was stopped
-	ServiceStopped StopStatus = iota
+	ServiceStopped StopStatus = iota + 10
```
kaidaguerre, line 25: comment to explain why +10

also if you come across code with this style of comment please update to 
```
ServiceStopped indicates service was stopped
```

db/local_db/status.go
```diff
@@ -36,3 +42,41 @@ func GetStatus() (*RunningDBInstanceInfo, error) {
...
+	}
+
+	// check if a process with that PID exists
+	exists, err := PidExists(int(pid))
+	if err != nil {
+		return err
+	}
+	if exists {
+		// if it does, then somehow we don't know about it. Error out
+		return fmt.Errorf("service is running in an unknown state [%d] - try killing it with %s", pid, constants.Bold("steampipe service stop --force"))
```
kaidaguerre, line 77: maybe 
```
service is running (PID: %d), but we have not state information  stored - try killing it with %s
```

db/local_db/status.go
```diff
@@ -36,3 +42,41 @@ func GetStatus() (*RunningDBInstanceInfo, error) {
...
+
+	// read the content of the postmaster.pid file
+	fileContent, err := ioutil.ReadFile(getPostmasterPidLocation())
+	if err != nil {
+		return err
+	}
+
+	// the first line contains the PID
+	lines := strings.FieldsFunc(string(fileContent), func(r rune) bool {
+		return r == '\n'
```
kaidaguerre, line 62: is this valid cross platform?

db/local_db/status.go
```diff
@@ -36,3 +42,41 @@ func GetStatus() (*RunningDBInstanceInfo, error) {
...
+	})
+	// extract it
+	pid, err := strconv.ParseInt(lines[0], 10, 64)
```
kaidaguerre, line 65: is `lines[0]` always non nil - fo we need to check `len(lines)`?

db/local_db/status.go
```diff
@@ -36,3 +42,41 @@ func GetStatus() (*RunningDBInstanceInfo, error) {
 
 	return info, nil
 }
+
+// errorIfUnknownService errors if it can find a `postmaster.pid` in the `INSTALL_DIR`
+// and the PID recorded in the found `postmaster.pid` is running
```
kaidaguerre, line 47: why does it error?

unclear from comment the reasoning here

kaidaguerre (changes requested):

db/local_db/status.go
binaek89: as per https://github.com/postgres/postgres/blob/76e38b37a5f179d4c9d2865ff31b79130407530b/src/bin/pg_ctl/pg_ctl.c#L317

`pg_ctl` uses `\n` to split lines in the `postmaster.pid` file

db/local_db/stop_database.go
binaek89: fixed

db/local_db/status.go
binaek89: fixed

db/local_db/status.go
binaek89: fixed

db/local_db/status.go
binaek89: updated comment

db/local_db/status.go
```diff
@@ -36,3 +42,52 @@ func GetStatus() (*RunningDBInstanceInfo, error) {
 
 	return info, nil
 }
+
+// errorIfUnknownService returns an error if it can find a `postmaster.pid` in the `INSTALL_DIR`
```
kaidaguerre, line 46: I'd still like a full explanation here - what is that file - why is it an error if the PID in the file is running?

PR #900, opened 2021-09-14
Adds support for outputting `stdout` and `stderr` from `service start` to `TRACE log`. Closes #810

### LINKED ISSUE #810: Trace log output of service if it fails to start.

binaek89:


db/local_db/start_database.go
```diff
@@ -211,6 +213,20 @@ func startPostgresProcess(port int, listen StartListenType, invoker constants.In
 		Foreground: false,
 	}
 
+	// create a channel with a big buffer, so that it doesn't choke
+	logChannel := make(chan string, 1000)
+	if stopListenFn, err := setupLogCollector(postgresCmd, logChannel); err == nil {
```
kaidaguerre, line 218: I think do not use the shortcut here, it's a little hard to parse - instead: 
```
stopListenFn, err := setupLogCollector(postgresCmd, logChannel)
if err == nil {
		defer func() {
			stopListenFn()
		}()
		go traceoutServiceLogs(logChannel)
	} else {
		// this is a convenience and therefore, we shouldn't error out if we
		// are not able to capture the logs.
		// instead, log to TRACE that we couldn't and continue
		log.Println("[TRACE] Warning: Could not attach to service logs")
	}```

binaek89: done

PR #923, opened 2021-09-17
Fix issue where `install.sh` fails for IPv6 enabled system. Closes #861

### LINKED ISSUE #861: `install.sh` installation fails when executed from an IPv6 enabled system
As per the suggested documentation in:
https://steampipe.io/downloads

When we execute:
```bash
sudo /bin/sh -c "$(curl -fsSL https://raw.githubusercontent.com/turbot/steampipe/main/install.sh)"
```

in a system which has `IPv6` enabled, the installation fails.

This may be an issue with `github.com` not supporting `IPv6` connections - as per an old community thread [here](https://github.community/t/cannot-reach-any-github-io-page-via-ipv6/10310/37)

binaek:


install.sh
```diff
@@ -58,11 +58,12 @@ echo "Downloading from $steampipe_uri"
 if command -v wget >/dev/null; then
 	# because --show-progress was introduced in 1.16.
 	wget --help | grep -q '\--showprogress' && _FORCE_PROGRESS_BAR="--no-verbose --show-progress" || _FORCE_PROGRESS_BAR=""
-	if ! wget --progress=bar:force:noscroll $_FORCE_PROGRESS_BAR -O "$zip_location" "$steampipe_uri"; then
+	if ! wget --prefer-family=IPv4 --progress=bar:force:noscroll $_FORCE_PROGRESS_BAR -O "$zip_location" "$steampipe_uri"; then
```
kaidaguerre, line 61: put in a comment explaining what we are doing

PR #935, opened 2021-09-21
Allow for setting service password from `ENV` or `cmd` flag. Closes #725

### LINKED ISSUE #725: Allow providing a user/pass combination to be used when running in Service Mode
**Current Issue:**
It is currently difficult to automate against a Steampipe Service instance with external tooling on the basis that it generates a password when started in service-mode that must be extracted from the stdout or logs.

This means that someone must manually acquire this information in order to configure other tools to interface with the psql endpoint - (or you'd have to write a script to parse the output but this seems clunky).

**Recommended Solution:**
I think the most automatable approach would be to add args to the command used when starting up Steampipe in service mode.

Example
`steampipe service start --user steampipe --pass s0m£SeCu|2€V4lu3!`

**Alternative Solutions:**
An alternative I've considered which may also be useful when using the container based Steampipe would be to store the desired username and password as `Environment Variables` or in a configuration file in the `.steampipe/` directory.

**Additional Context:**
The main reason for requesting this is that I would like to be able to automate a deployment of Steampipe and a few external tools which will interface with Steampipe via the psql endpoint created by it's service mode.

At the current time this requires the following steps:
- Deploy Steampipe in service mode and manually obtain credentials.
- Place credentials into secret engine (manually)
- Deploy other tooling using the information stored.

I would rather have all this automated so I can generate the credentials and deploy all of the services (Steampipe, etc) together all nicely automated.

binaek:
This PR adds support for updating the `service password` for the `steampipe` user.

Password can be set through:
* `environment` variable `STEAMPIPE_DATABASE_PASSWORD`
* `--database-password` flag in `steampipe service start`

The `--database-password` take precedence over `STEAMPIPE_DATABASE_PASSWORD` environment variable.

If neither is provided, then the service defaults to a password which is generated during installation in `$STEAMPIPE_INSTALL_DIR/internal/.passwd`.

During `steampipe service start`, if the `$STEAMPIPE_INSTALL_DIR/internal/.passwd` is not found, then one is generated with a `random` string.

`STEAMPIPE_DATABASE_PASSWORD` environment is effective in all commands which start `service` implicitly, if required.

> Note: if the service is already started, then `STEAMPIPE_DATABASE_PASSWORD` has no effect and will be ignored.


db/db_local/install.go
```diff
@@ -238,18 +238,17 @@ func doInit(firstInstall bool, spinner *spinner.Spinner) error {
 
 	display.UpdateSpinnerMessage(spinner, "Generating database passwords...")
 	// Try for passwords of the form dbC-3Ji-d04d
```
kaidaguerre, line 240: is this comment still correct?

db/db_local/password.go
```diff
@@ -14,30 +17,16 @@ type Passwords struct {
...
 
-func writePasswordFile(steampipePassword string, rootPassword string) error {
-	log.Println("[TRACE]", "Writing password file")
-	passwords := Passwords{
-		Root:      rootPassword,
-		Steampipe: steampipePassword,
+func readPassword() (string, error) {
+	if !helpers.FileExists(getPasswordFileLocation()) {
+		p := generatePassword()
+		return p, ioutil.WriteFile(getPasswordFileLocation(), []byte(p), 0600)
```
kaidaguerre, line 23: let's split this into a couple of lines for clarity

kaidaguerre:
```
if err := writePasswordFile(p); err != nil{
   return "", err
}
return p, nil
```

db/db_local/password.go
```diff
@@ -50,5 +39,32 @@ func generatePassword() string {
...
+func migrateLegacyPasswordFile() error {
+	utils.LogTime("db_local.migrateLegacyPasswordFile start")
+	defer utils.LogTime("db_local.migrateLegacyPasswordFile end")
+	if helpers.FileExists(getLegacyPasswordFileLocation()) {
+		p, err := getLegacyPasswords()
+		if err != nil {
+			return err
+		}
+		os.Remove(getLegacyPasswordFileLocation())
+		return ioutil.WriteFile(getPasswordFileLocation(), []byte(p.Steampipe), 0600)
```
kaidaguerre, line 54:
```
return  writePasswordFile(p.Steampipe)
```

db/db_local/start_database.go
```diff
@@ -226,27 +246,31 @@ func startPostgresProcess(port int, listen StartListenType, invoker constants.In
...
 	}
 
-	// get the password file
-	passwords, err := getPasswords()
+	// get the password from the password file
+	password, err := readPassword()
 	if err != nil {
 		return err
 	}
+	if viper.IsSet(constants.ArgServicePassword) {
```
kaidaguerre, line 254: comment - what are we doing here?

db/db_local/start_database.go
```diff
@@ -161,6 +156,31 @@ func startPostgresProcess(port int, listen StartListenType, invoker constants.In
 	if err != nil {
 		return err
 	}
+	return nil
+}
+
+// startPostgresProcessAndSetPassword starts the postgres process and writes out the state file
+// after it is convinced that the process is started and is accepting connections
+func startPostgresProcessAndSetPassword(port int, listen StartListenType, invoker constants.Invoker) (e error) {
```
kaidaguerre, line 164: this function is way too big
move 
```
postgresCmd := exec.Command(
		getPostgresBinaryExecutablePath(),
		// by this time, we are sure that the port if free to listen to
		"-p", fmt.Sprint(port),
		"-c", fmt.Sprintf("listen_addresses=\"%s\"", listenAddresses),
		// NOTE: If quoted, the application name includes the quotes. Worried about
		// having spaces in the APPNAME, but leaving it unquoted since currently
		// the APPNAME is hardcoded to be steampipe.
		"-c", fmt.Sprintf("application_name=%s", constants.APPNAME),
		"-c", fmt.Sprintf("cluster_name=%s", constants.APPNAME),

		// log directory
		"-c", fmt.Sprintf("log_directory=%s", constants.LogDir()),

		// If ssl is off  it doesnot matter what we pass in the ssl_cert_file and ssl_key_file
		// SSL will only get validated if the ssl is on
		"-c", fmt.Sprintf("ssl=%s", SslStatus()),
		"-c", fmt.Sprintf("ssl_cert_file=%s", filepath.Join(getDataLocation(), constants.ServerCert)),
		"-c", fmt.Sprintf("ssl_key_file=%s", filepath.Join(getDataLocation(), constants.ServerKey)),

		// Data Directory
		"-D", getDataLocation())

	postgresCmd.Env = append(os.Environ(), fmt.Sprintf("STEAMPIPE_INSTALL_DIR=%s", constants.SteampipeDir))
``` 
into a function

and also
```
// create a channel with a big buffer, so that it doesn't choke
	logChannel := make(chan string, 1000)
	stopListenFn, err := setupLogCollector(postgresCmd, logChannel)
	if err == nil {
		defer func() {
			stopListenFn()
		}()
		go traceoutServiceLogs(logChannel)
	} else {
		// this is a convenience and therefore, we shouldn't error out if we
		// are not able to capture the logs.
		// instead, log to TRACE that we couldn't and continue
		log.Println("[TRACE] Warning: Could not attach to service logs")
	}
```

kaidaguerre (changes requested):

db/db_local/password.go
binaek: split out

db/db_local/start_database.go
```diff
@@ -134,34 +140,119 @@ func StartDB(port int, listen StartListenType, invoker constants.Invoker) (start
...
-	// create the postgresql.conf.d location, don't fail if it errors
-	err = os.MkdirAll(getPostgresqlConfDLocation(), 0700)
+	// if a password was set through the `STEAMPIPE_DATABASE_PASSWORD` environment variable
+	// or through the `--database-password` cmdline flag, then use that for this session
+	// instead of the default one
+	if viper.IsSet(constants.ArgServicePassword) {
+		password = viper.GetString(constants.ArgServicePassword)
+	}
+
+	runningInfo := new(RunningDBInstanceInfo)
```
kaidaguerre, line 205: `createRunningInfo()`

db/db_local/password.go
```diff
@@ -14,30 +17,23 @@ type Passwords struct {
...
+func writePasswordFile(password string) error {
+	return ioutil.WriteFile(getPasswordFileLocation(), []byte(password), 0600)
 }
 
-func getPasswords() (*Passwords, error) {
-	contentBytes, err := ioutil.ReadFile(getPasswordFileLocation())
-	if err != nil {
-		return nil, err
+func readPasswordFile() (string, error) {
+	if !helpers.FileExists(getPasswordFileLocation()) {
```
kaidaguerre, line 25: add comment

kaidaguerre (changes requested):

db/db_local/start_database.go
```diff
@@ -27,7 +28,8 @@ type StartListenType string
 
 const (
 	// ServiceStarted :: StartResult - Service was started
-	ServiceStarted StartResult = iota
+	// start from 10 to prevent confusion with int zero-value
+	ServiceStarted StartResult = iota + 10
```
kaidaguerre, line 32: can't you just do 
```
ServiceStarted StartResult iota =1
```

db/db_local/start_database.go
```diff
@@ -134,18 +140,85 @@ func StartDB(port int, listen StartListenType, invoker constants.Invoker) (start
...
+
+	err = setupServicePassword(invoker, password)
+	if err != nil {
+		postgresCmd.Process.Kill()
+		return err
+	}
+	return nil
+}
+
+func setupServicePassword(invoker constants.Invoker, password string) error {
```
kaidaguerre, line 206: can you ensure these functions appear in the order they are called in startPostgresProcessAndSetPassword

db/db_local/stop_database.go
```diff
@@ -89,19 +89,19 @@ func GetCountOfConnectedClients() (int, error) {
 }
 
 // StopDB searches for and stops the running instance. Does nothing if an instance was not found
-func StopDB(force bool, invoker constants.Invoker, spinner *spinner.Spinner) (StopStatus, error) {
+func StopDB(force bool, invoker constants.Invoker, spinner *spinner.Spinner) (status StopStatus, e error) {
 	log.Println("[TRACE] StopDB", force)
```
kaidaguerre, line 94: remove line

kaidaguerre (changes requested):

db/db_local/start_database.go
binaek: Unfortunately , this doesn't compile

kaidaguerre: in which case let's use iota+1

kaidaguerre (changes requested):

PR #1087, opened 2021-11-08
Fixes issue where service is not shutdown when a `check` is cancelled. Closes 1081

binaek:


db/db_client/db_client_execute.go
```diff
@@ -93,8 +90,8 @@ func (c *DbClient) ExecuteInSession(ctx context.Context, session *sql.Conn, quer
 			}
 			// TODO change this to NOT close the session here - instead maybe call a passed-in OnComplete callback
```
kaidaguerre, line 91: remove comment

db/db_client/db_client.go
```diff
@@ -39,9 +40,10 @@ func NewDbClient(connectionString string) (*DbClient, error) {
 	client := &DbClient{
 		dbClient:            db,
 		sessionAcquireLock:  new(sync.Mutex),
-		initializedSessions: make(map[int64]SessionStats),
+		initializedSessions: make(map[int64]*SessionStats),
 		// set up a blank struct for the schema metadata
-		schemaMetadata: schema.NewMetadata(),
+		schemaMetadata:       schema.NewMetadata(),
+		sessionInitWaitGroup: &sync.WaitGroup{},
```
kaidaguerre, line 46: comment - whyt do we need this

kaidaguerre (changes requested):

PR #1099, opened 2021-11-10
Fix issue where pg `usernames` weren't getting escaped properly when setting `search path`. Closes 1094. Closes 1082.

binaek:


db/db_local/local_db_client.go
```diff
@@ -221,7 +221,7 @@ func (c *LocalDbClient) setUserSearchPath() ([]string, error) {
 		}
 		queries = append(queries, fmt.Sprintf(
 			"alter user %s set search_path to %s;",
-			user,
+			db_common.PgEscapeName(user),
 			strings.Join(escapedSearchPath, ","),
 		))
```
kaidaguerre, line 226: can we have an acceptance test

kaidaguerre (changes requested):

PR #1100, opened 2021-11-10
Add support to fall back to `more` (if available) - where `less` is not available in the environment. Closes #1072

### LINKED ISSUE #1072: Interactive query displays no results on Debian 9.13
**Describe the bug**
Running `steampipe query` interactively, no results are displayed when the output mode is `table`.  This is due to a dependence on `less`, which is not installed in Debian 9.13.  Installing `less` resolves the issue.  Ideally, steampipe should not rely on `less`, but if it does then we should at least print an informative help message.

steampipe version 0.9.0 (but likely ALL versions)

**To reproduce**
Install steampipe on a system that dies not have 'less' installed (debian 9.13 for example).  Run `steampipe query` interactively, and run a query.

**Expected behavior**
Ideally, steampipe should not rely on `less`, but if it does then we should at least print an informative help message.

**Additional context**

binaek:


display/pager.go
```diff
@@ -11,17 +11,15 @@ import (
...
 // ShowPaged :: displays the `content` in a system dependent pager
 func ShowPaged(content string) {
 	if isPagerNeeded(content) {
-		switch runtime.GOOS {
-		case "darwin", "linux":
-			nixLessPager(content)
-		case "windows":
-			winMorePager(content)
-		default:
+		if (runtime.GOOS == "darwin" || runtime.GOOS == "linux") && (isMoreAvailable() || isLessAvailable()) {
```
kaidaguerre, line 20:
```
if (runtime.GOOS == "darwin" || runtime.GOOS == "linux")
			nixPager(content)
		} 
```

display/pager.go
```diff
@@ -59,24 +57,39 @@ func isPagerNeeded(content string) bool {
...
 	// we will use this for non-tty environments as well
 	fmt.Print(content)
 }
 
-func nixLessPager(content string) {
-	lessProcess := exec.Command("less", "-SRXF")
-	lessProcess.Stdout = os.Stdout
-	lessProcess.Stderr = os.Stderr
-	lessProcess.Stdin = strings.NewReader(content)
+func nixMoreOrLessPager(content string) {
```
kaidaguerre, line 66:
```
func nixPager(content string) {
	if isLessAvailable() {
		pager(exec.Command("less", "-SRXF"), content)
	} else if isMoreAvailable() {
		pager(exec.Command("more"), content)
	} else {
		nullPager(content)
	}
}

func pager(cmd *exec.Cmd){}
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	cmd.Stdin = strings.NewReader(content)
	// Run it, so that this blocks out the go-prompt stuff.
	// No point Start-ing it anyway
	err := cmd.Run()
	if err != nil {
		utils.ShowErrorWithMessage(err, "could not display results")
	}
}
```

kaidaguerre (changes requested):

display/pager.go
binaek: done

PR #1185, opened 2021-11-25
Brings stability to `service stop`. Closes #1180

### LINKED ISSUE #1180: Stabilize `service stop`
`service stop` sometimes does not stop the service and leaves it in a bad state.

This is evident from the message:

`Error: service is running in an unknown state [PID: xxxxx] - try killing it with steampipe service stop --force`

which comes up from time-to-time

binaek:


binaek:
experimental PR

PR #1220, opened 2021-12-10
Fix issue where running `service`s had to be stopped to install in a new `install-dir`

binaek:


db/db_local/install.go
```diff
@@ -232,13 +230,18 @@ func runInstall(ctx context.Context, firstInstall bool, spinner *spinner.Spinner
 		log.Printf("[TRACE] getNextFreePort failed: %v", err)
 		return fmt.Errorf("Starting database... FAILED!")
 	}
+
+	time.Sleep(5 * time.Second)
```
kaidaguerre, line 234: sleeps????

kaidaguerre (changes requested):

db/db_local/install.go
binaek: removed

PR #1276, opened 2022-01-03
Removes transaction creation around query execution. Closes #1241

### LINKED ISSUE #1241: Transaction should not be created for queries.

binaek:


db/db_client/db_client_execute.go
```diff
@@ -75,6 +75,10 @@ func (c *DbClient) Execute(ctx context.Context, query string, disableSpinner boo
 	return c.ExecuteInSession(ctx, sessionResult.Session, query, closeSessionCallback, disableSpinner)
 }
 
+// ExecuteInSession implements Client
+// execute the provided query against the Service in the given `context.Context` using the provided `*DatabaseSession`
```
kaidaguerre, line 79: no slanted quotes in comments 
```
// ExecuteInSession implements Client
// execute the query in the given Context using the provided DatabaseSession
// ExecuteInSession assumes no responsibility over the lifecycle of the DatabaseSession - that is the responsibility of the caller
// NOTE: The returned Result MUST be fully read - otherwise the connection will block and will prevent further communication

File: db/db_client/db_client_transaction.go [hunk 2]
```diff
@@ -13,6 +13,9 @@ import (
 // createTransaction , with a timeout - this may be required if the db client becomes unresponsive
 func (c *DbClient) createTransaction(ctx context.Context, session *sql.Conn, retryOnTimeout bool) (tx *sql.Tx, err error) {
 	doneChan := make(chan bool)
+	if session == nil {
+		return nil, fmt.Errorf("cannot create transaction :: received nil SESSION")
```
kaidaguerre, line 17: do we still use this function?

kaidaguerre (changes requested):

db/db_client/db_client_transaction.go
binaek: removed

PR #1280, opened 2022-01-03
Fix issue where service is not shutdown if interrupted while interactive prompt is initialising. Closes #1004

### LINKED ISSUE #1004: In interactive mode, if ctrl+d is hit before initialisation is complete, the database is not shut down

binaek:


interactive/interactive_client.go
```diff
@@ -99,10 +96,9 @@ func (c *InteractiveClient) InteractivePrompt() {
 		// close up the SIGINT channel so that the receiver goroutine can quit
 		signal.Stop(interruptSignalChannel)
 		close(interruptSignalChannel)
-		// close the underlying client if it exists
-		if client := c.client(); client != nil {
-			client.Close()
-		}
+
+		// init data cleanup
```
kaidaguerre, line 100: need a better comment here - this is important - this cleanup shuts the service
```
// cleanup the init data to ensure any services we started are stopped
```

interactive/interactive_client_init.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/turbot/steampipe/db/db_common"
+	"github.com/turbot/steampipe/workspace"
```
kaidaguerre, line 11: remove gap

kaidaguerre (changes requested):
looks good - minor tweaks

interactive/interactive_client.go
binaek: 👍

interactive/interactive_client_init.go
binaek: done

PR #1361, opened 2022-01-27
Fix issues reported by the 'unused', 'staticcheck', 'gosimple' linters. closes #1360

### LINKED ISSUE #1360: Fix issues reported by the 'unused' and 'gosimple' linters

pskrbasu:
@kaidaguerre Is this function needed for reporting?
```
cmd/report.go:14:6: func `runReportCmd` is unused (unused)
func runReportCmd(cmd *cobra.Command, args []string) {
     ^
```

### DIFF
--- a/query/metaquery/completers.go
+++ b/query/metaquery/completers.go
@@ -31,7 +31,7 @@ func Complete(input *CompleterInput) []prompt.Suggest {
 
 func completerFromArgsOf(cmd string) completer {
 	return func(input *CompleterInput) []prompt.Suggest {
-		metaQueryDefinition, _ := metaQueryDefinitions[cmd]
+		metaQueryDefinition := metaQueryDefinitions[cmd]
 		suggestions := make([]prompt.Suggest, len(metaQueryDefinition.args))
 		for idx, arg := range metaQueryDefinition.args {
 			suggestions[idx] = prompt.Suggest{Text: arg.value, Description: arg.description, Output: arg.value}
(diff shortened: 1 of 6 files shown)


kaidaguerre:
yup - that function is used in the (soon to be merged)reporting branch

query/metaquery/completers.go
```diff
@@ -31,7 +31,7 @@ func Complete(input *CompleterInput) []prompt.Suggest {
 
 func completerFromArgsOf(cmd string) completer {
 	return func(input *CompleterInput) []prompt.Suggest {
-		metaQueryDefinition, _ := metaQueryDefinitions[cmd]
+		metaQueryDefinition := metaQueryDefinitions[cmd]
```
e-gineer, line 34: Is this a change that continues to ignore an error? If so, shouldn't we handle the error?

pskrbasu: `metaQueryDefinitions` is a map[string]metaQueryDefinition. We are just assigning the value here.

kaidaguerre: to expand: in this map syntax ```metaQueryDefinition, _ := metaQueryDefinitions[cmd]``` the second return val is a boolean indicating whether the value was found

steampipeconfig/parse/decode.go
```diff
@@ -287,7 +288,8 @@ func decodeParam(block *hcl.Block, runCtx *RunContext, parentName string) (*modc
 		diags = append(diags, valDiags...)
 	}
 	if attr, exists := content.Attributes["default"]; exists {
-		v, diags := attr.Expr.Value(runCtx.EvalCtx)
+		var v cty.Value
```
kaidaguerre, line 291: what is this change for?

PR #1367, opened 2022-01-28
Integrate reporting server assets. Closes #1364

### LINKED ISSUE #1364: Integrate Reports UI

binaek:


constants/assets.go
```diff
@@ -0,0 +1,6 @@
+package constants
+
```
kaidaguerre, line 2: do we need this and the constants defined in report.go?

kaidaguerre (changes requested):

PR #1377, opened 2022-02-01
Create and use a `$STEAMPIPE_INSTALL_DIR/tmp` for `oci` installations. Closes #1375

### LINKED ISSUE #1375: hub download should use dedicated `tmpdir` instead of `plugin` directory

binaek:


filepaths/steampipe.go
```diff
@@ -19,53 +19,58 @@ const (
...
-	if _, err := os.Stat(subDir); os.IsNotExist(err) {
+	if _, err := os.Stat(subDir); os.IsNotExist(err) && createOnNotExist {
 		err = os.MkdirAll(subDir, 0755)
 		utils.FailOnErrorWithMessage(err, fmt.Sprintf("could not create %s directory", dirName))
 	}
 
 	return subDir
 }
 
+// TmpDir returns the path to the tmp directory in STEAMPIPE_HOME (creates if missing)
```
kaidaguerre, line 36: OK lets rename all the dir functions `Ensurexxxxx`, e.,g. `EnsureTemplateDir`

It's a bit unexpected that calling what seems like a simple name function creates the directory - I've never quite liked that

Then we can have the functions `TmpDir` and `EnsureTmpDir`

kaidaguerre (changes requested):

PR #1502, opened 2022-02-23
Fix issue where schema extraction is broken for remote services. Closes #1497

### LINKED ISSUE #1497: Schema extraction is broken for remote databases
When extracting connections, `steampipe` looks up the local config

This approach WILL not work for remote connections

Schema extraction should be independent of local config

binaek:


db/db_local/local_db_client.go
```diff
@@ -154,8 +154,90 @@ func (c *LocalDbClient) ContructSearchPath(ctx context.Context, requiredSearchPa
 	return c.client.ContructSearchPath(ctx, requiredSearchPath, searchPathPrefix, currentSearchPath)
 }
 
-func (c *LocalDbClient) GetSchemaFromDB(ctx context.Context, schemas []string) (*schema.Metadata, error) {
-	return c.client.GetSchemaFromDB(ctx, schemas)
+func (c *LocalDbClient) GetSchemaFromDB(ctx context.Context) (*schema.Metadata, error) {
```
kaidaguerre, line 157: add comment to explain why local  db has a different implementation

db/db_local/local_db_client.go
```diff
@@ -154,8 +154,90 @@ func (c *LocalDbClient) ContructSearchPath(ctx context.Context, requiredSearchPa
...
+	// (pass nil for connection state - this forces NewConnectionSchemaMap to load it)
+	connectionSchemaMap, err := steampipeconfig.NewConnectionSchemaMap()
+	if err != nil {
+		return nil, err
+	}
+	// get the unique schema - we use this to limit the schemas we load from the database
+	schemas := connectionSchemaMap.UniqueSchemas()
+	query := c.buildSchemasQuery(schemas)
+
+	acquireSessionResult := c.AcquireSession(ctx)
```
kaidaguerre, line 168: why do we need to acquire session here, but the base implementation just calls 

```
connection, err := c.dbClient.Conn(ctx)
```

db/db_local/local_db_client.go
```diff
@@ -154,8 +154,90 @@ func (c *LocalDbClient) ContructSearchPath(ctx context.Context, requiredSearchPa
...
+	searchPath, err := c.GetCurrentSearchPath(ctx)
+	if err != nil {
+		return nil, err
+	}
+	metadata.SearchPath = searchPath
+
+	return metadata, nil
+}
+
+func (c *LocalDbClient) populateSchemaMetadata(schemaMetadata *schema.Metadata, connectionSchemaMap steampipeconfig.ConnectionSchemaMap) {
```
kaidaguerre, line 197: add function comment
```
// update schemaMetadata to add in all other schemas which have the same schemas as those we have loaded
// NOTE: this mutates 	schemaMetadata
```

db/db_local/local_db_client.go
```diff
@@ -154,8 +154,90 @@ func (c *LocalDbClient) ContructSearchPath(ctx context.Context, requiredSearchPa
...
+			exemplarSchema = make(map[string]schema.TableSchema)
+		}
+
+		for _, s := range otherSchemas {
+			schemaMetadata.Schemas[s] = exemplarSchema
+		}
+	}
+}
+
+func (c *LocalDbClient) buildSchemasQuery(hintSchemas []string) string {
```
kaidaguerre, line 213: what are hintSchemas?

just call it `schemas`

db/db_local/local_db_client.go
```diff
@@ -154,8 +154,90 @@ func (c *LocalDbClient) ContructSearchPath(ctx context.Context, requiredSearchPa
...
+			// should can happen in the case of a dynamic plugin with no tables - use empty schema
+			exemplarSchema = make(map[string]schema.TableSchema)
+		}
+
+		for _, s := range otherSchemas {
+			schemaMetadata.Schemas[s] = exemplarSchema
+		}
+	}
+}
+
```
kaidaguerre, line 212: function comment to explain why this is different from the base version

kaidaguerre (changes requested):

db/db_local/local_db_client.go
binaek: This is because `dbClient` is not a public property in the `DbClient` struct

db/db_local/local_db_client.go
```diff
@@ -194,6 +196,8 @@ func (c *LocalDbClient) GetSchemaFromDB(ctx context.Context) (*schema.Metadata,
 	return metadata, nil
 }
 
+// update schemaMetadata to add in all other schemas which have the same schemas as those we have loaded
+// NOTE: this mutates 	schemaMetadata
```
kaidaguerre, line 200: feel free to remove the extra spaces from my suggested comment

db/db_local/local_db_client.go
```diff
@@ -154,8 +154,94 @@ func (c *LocalDbClient) ContructSearchPath(ctx context.Context, requiredSearchPa
...
+			exemplarSchema = make(map[string]schema.TableSchema)
+		}
+
+		for _, s := range otherSchemas {
+			schemaMetadata.Schemas[s] = exemplarSchema
+		}
+	}
+}
+
+func (c *LocalDbClient) buildSchemasQuery(exemplarSchemas []string) string {
```
kaidaguerre, line 217: just `schemas` will be fine

kaidaguerre (changes requested):

PR #1678, opened 2022-03-16
Fixes issue where installing a 'non-latest' plugin will not install config even if it's not there. Closes #1660

### LINKED ISSUE #1660: Update default config file during plugin install.
If the plugin getting installed is not the `latest` tagged version, then the `config.spc` in the plugin artifact needs to be modified to have the exact tag in the `plugin` property before writing to the `config` directory.

binaek:


ociinstaller/plugin.go
```diff
@@ -126,18 +128,60 @@ func installPluginConfigFiles(image *SteampipeImage, tempdir string) error {
...
+	substitution := fmt.Sprintf(`$1 plugin = "$2@%s"`, stream)
+
+	srcScanner := bufio.NewScanner(strings.NewReader(string(src)))
+	srcScanner.Split(bufio.ScanLines)
+	destBuffer := bytes.NewBufferString("")
+
+	for srcScanner.Scan() {
+		line := srcScanner.Text()
+		if regex.MatchString(line) {
+			fmt.Println("match with", line)
```
kaidaguerre, line 176: what is this printLn for?

ociinstaller/plugin.go
```diff
@@ -126,18 +128,60 @@ func installPluginConfigFiles(image *SteampipeImage, tempdir string) error {
...
+			line = regex.ReplaceAllString(line, substitution)
+		}
```
kaidaguerre, line 178: once we have found a match, can we stop scanning?

ociinstaller/plugin.go
```diff
@@ -126,18 +128,60 @@ func installPluginConfigFiles(image *SteampipeImage, tempdir string) error {
...
 
+	if err = os.WriteFile(destFile, transformedData, inputStat.Mode()); err != nil {
+		return fmt.Errorf("writing to output file failed: %s", err)
+	}
 	return nil
 }
 
-func pluginInstallDir(imageRef string) string {
-	ref := NewSteampipeImageRef(imageRef)
+func transform(src []byte, ref *SteampipeImageRef) []byte {
```
kaidaguerre, line 160: give a more informative name - renamePlugin

ociinstaller/plugin.go
```diff
@@ -126,18 +128,60 @@ func installPluginConfigFiles(image *SteampipeImage, tempdir string) error {
...
+	_, _, stream := ref.GetOrgNameAndStream()
+	if stream == "latest" {
```
kaidaguerre, line 162: move this check to outside the function - then the renamed function name will make sense

ociinstaller/plugin.go
kaidaguerre: only better than that :)

kaidaguerre (changes requested):

ociinstaller/plugin.go
```diff
@@ -149,20 +149,18 @@ func copyConfigFileUnlessExists(sourceFile string, destFile string, ref *Steampi
 		return fmt.Errorf("couldn't read source file permissions: %s", err)
 	}
 	// transform
```
kaidaguerre, line 151: better comment

ociinstaller/plugin.go
```diff
@@ -149,20 +149,18 @@ func copyConfigFileUnlessExists(sourceFile string, destFile string, ref *Steampi
...
-
+	inputData = addPluginStreamToConfig(inputData, ref, "latest")
 	if err = os.WriteFile(destFile, inputData, inputStat.Mode()); err != nil {
 		return fmt.Errorf("writing to output file failed: %s", err)
 	}
 	return nil
 }
 
-func addPluginStreamToConfig(src []byte, ref *SteampipeImageRef) []byte {
+func addPluginStreamToConfig(src []byte, ref *SteampipeImageRef, nopIfStream string) []byte {
```
kaidaguerre, line 159: comment - what does this function do?

explain what and why

ociinstaller/plugin.go
```diff
@@ -149,20 +149,18 @@ func copyConfigFileUnlessExists(sourceFile string, destFile string, ref *Steampi
...
 	_, _, stream := ref.GetOrgNameAndStream()
+	if stream == nopIfStream {
```
kaidaguerre, line 161: `nopIfStream`????

ociinstaller/plugin.go
```diff
@@ -126,18 +128,59 @@ func installPluginConfigFiles(image *SteampipeImage, tempdir string) error {
...
+
+	srcScanner := bufio.NewScanner(strings.NewReader(string(src)))
+	srcScanner.Split(bufio.ScanLines)
+	destBuffer := bytes.NewBufferString("")
+
+	for srcScanner.Scan() {
+		line := srcScanner.Text()
+		if regex.MatchString(line) {
+			line = regex.ReplaceAllString(line, substitution)
+			line = line[1:]
```
kaidaguerre, line 176: why?

kaidaguerre (changes requested):

ociinstaller/plugin.go
```diff
@@ -148,17 +148,21 @@ func copyConfigFileUnlessExists(sourceFile string, destFile string, ref *Steampi
 	if err != nil {
 		return fmt.Errorf("couldn't read source file permissions: %s", err)
 	}
-	// transform
-	inputData = addPluginStreamToConfig(inputData, ref, "latest")
+	inputData = addPluginStreamToConfig(inputData, ref)
```
kaidaguerre, line 151: // update the connection config with the correct plugin version

ociinstaller/plugin.go
```diff
@@ -148,17 +148,21 @@ func copyConfigFileUnlessExists(sourceFile string, destFile string, ref *Steampi
...
 	if err = os.WriteFile(destFile, inputData, inputStat.Mode()); err != nil {
 		return fmt.Errorf("writing to output file failed: %s", err)
 	}
 	return nil
 }
 
-func addPluginStreamToConfig(src []byte, ref *SteampipeImageRef, nopIfStream string) []byte {
+// The default config files have the plugin set to the latest stream
```
kaidaguerre, line 158:
```
// The default config files have the plugin set to the 'latest' stream (as this is what is installed by default)
// When installing non-latest plugins, that property needs to be adjusted to the stream actually getting installed. 
// Otherwise, during plugin resolution, it will resolve to an incorrect plugin instance
// (or none at at all, if  'latest' versions isn't installed)

kaidaguerre (changes requested):

ociinstaller/plugin.go
binaek: This is because in the substitution `regex` I had to put in an extra space to isolate the substitution token `$1`. Removing it here.

kaidaguerre (changes requested):

PR #1793, opened 2022-04-04
Fix issue where plugin install always install amd64 for plugins.

binaek:


ociinstaller/steampipeimage.go
```diff
@@ -186,12 +186,13 @@ func getPluginImageData(layers []ocispec.Descriptor) (*PluginImage, error) {
...
-			log.Println("[TRACE] falling back to the next one, if any")
-			continue
+		if len(foundLayers) == 1 {
+			res.BinaryFile = foundLayers[0].Annotations["org.opencontainers.image.title"]
+			break
 		}
-		res.BinaryFile = foundLayers[0].Annotations["org.opencontainers.image.title"]
+		log.Println("[TRACE] could not find data for", mediaType)
+		log.Println("[TRACE] falling back to the next one, if any")
+		continue
```
kaidaguerre, line 195: continue is not needed

ociinstaller/steampipeimage.go
```diff
@@ -186,12 +186,13 @@ func getPluginImageData(layers []ocispec.Descriptor) (*PluginImage, error) {
 	// get the binary plugin file info
 	for _, mediaType := range MediaTypeForPlatform("plugin") {
```
kaidaguerre, line 187: this code is a bit opaque - at least out of context like this

Add a short comment to explain thew control flow

kaidaguerre (changes requested):

ociinstaller/steampipeimage.go
binaek: done

ociinstaller/steampipeimage.go
binaek: :-D

PR #1856, opened 2022-04-13
Add support for plugins to switch over to arm64 build for darwin even in the same version.

binaek:


binaek:
overridden by #1858

PR #1858, opened 2022-04-13
Add support for plugins to switch over to arm64 build for darwin even in the same version. Closes #1796

### LINKED ISSUE #1796: In `darwin/arm64` plugins should switch to `arm64` build, if available, even if an `amd64` of the same version is installed

binaek:


plugin/update.go
```diff
@@ -0,0 +1,29 @@
...
+func SkipUpdate(report VersionCheckReport) (bool, string) {
+	if report.Plugin.ImageDigest == report.CheckResponse.Digest {
+		return true, constants.PluginLatestAlreadyInstalled
+	}
+	if isMacM1() && len(report.Plugin.BinaryArchitecture) == 0 && hasM1Binary(report.CheckResponse.Manifest) {
+		return false, ""
+	}
+	return false, ""
+}
+func isMacM1() bool {
```
kaidaguerre, line 19: newline between functions

plugin/update.go
```diff
@@ -0,0 +1,29 @@
+package plugin
+
+import (
+	"runtime"
+
+	"github.com/turbot/steampipe/constants"
+	"github.com/turbot/steampipe/ociinstaller"
+)
+
+func SkipUpdate(report VersionCheckReport) (bool, string) {
```
kaidaguerre, line 10: function comment - explain the algorithm

this function is the core of the logic so make it super clear

plugin/version_checker.go
```diff
@@ -138,19 +129,21 @@ func (v *VersionChecker) getLatestVersionsForPlugins(plugins []*versionfile.Inst
 		return map[string]VersionCheckReport{}
 	}
 
+	log.Println("[TRACE] serverResponse:", serverResponse)
+
 	for _, rD := range serverResponse {
```
kaidaguerre, line 134: rD?????

kaidaguerre (changes requested):

plugin/update.go
kaidaguerre: // 1) If there is an updated version ALWAYS update
	// 2) If we are ARM, current installed version is AMD, and ARM is available
	// 3) Otherwise skip

plugin/update.go
kaidaguerre, line 19: add comment - returns whether RUNNING as M1

kaidaguerre (changes requested):

ociinstaller/mediatypes.go
```diff
@@ -52,18 +52,17 @@ const (
 // and it's fallbacks in order of priority
 func MediaTypeForPlatform(imageType ImageType) []string {
 	arch := runtime.GOARCH
-	layerFmt := "application/vnd.turbot.steampipe.%s.%s-%s.layer.v1+tar"
 	switch imageType {
 	case ImageTypeDatabase:
-		return []string{fmt.Sprintf(layerFmt, imageType, runtime.GOOS, arch)}
+		return []string{fmt.Sprintf("application/vnd.turbot.steampipe.%s.%s-%s.layer.v1+tar", imageType, runtime.GOOS, arch)}
```
kaidaguerre, line 57: can we define `layerFmtGzip` and  `layerFmtTar`

kaidaguerre (changes requested):

ociinstaller/mediatypes.go
binaek: i removed this

PR #1881, opened 2022-04-18
Fix the in-place switch over to ARM for plugins . Closes #1880

### LINKED ISSUE #1880: Fix the in-place switch over to ARM for plugins

pskrbasu:


task/pluginversioncheck.go
```diff
@@ -63,13 +63,17 @@ func pluginNotificationMessage(reports []plugin.VersionCheckReport) []string {
 				constants.Bold(report.CheckResponse.Version),
 			)
 		} else {
+			version := report.CheckResponse.Version
 			format := fmt.Sprintf("  %%-%ds @ %%-10s       %%10s → %%-10s", longestNameLength)
+			if report.Plugin.Version == report.CheckResponse.Version {
+				version = fmt.Sprintf("%s (arm64)", version)
```
kaidaguerre, line 69: add a comment  - why are we doing this?

plugin/update.go
```diff
@@ -17,7 +17,7 @@ func SkipUpdate(report VersionCheckReport) (bool, string) {
 	}
 
 	// 2) If we are M1, current installed version is AMD, and ARM is available - update
-	if isRunningAsMacM1() && report.Plugin.BinaryArchitecture == constants.ArchARM64 && manifestHasM1Binary(report.CheckResponse.Manifest) {
+	if isRunningAsMacM1() && manifestHasM1Binary(report.CheckResponse.Manifest) && report.Plugin.BinaryArchitecture != constants.ArchARM64 {
```
kaidaguerre, line 20: reverse logic of SkipUpdate and call it ShouldUpdate

negative bools are always a bit confusing

kaidaguerre (changes requested):

PR #1892, opened 2022-04-19
Detect the underlying arch to install the correct FDW package. Closes #1889

### LINKED ISSUE #1889: Detect the underlying arch to install the correct FDW package

pskrbasu:


ociinstaller/mediatypes.go
```diff
@@ -59,7 +60,7 @@ func MediaTypeForPlatform(imageType ImageType) []string {
 	case ImageTypeDatabase:
 		return []string{fmt.Sprintf(layerFmtTar, imageType, runtime.GOOS, arch)}
 	case ImageTypeFdw:
-		return []string{fmt.Sprintf(layerFmtGzip, imageType, runtime.GOOS, arch)}
+		return []string{fmt.Sprintf(layerFmtGzip, imageType, runtime.GOOS, utils.UnderlyingArch())} // Detect the underlying architecture(amd64/arm64)
```
kaidaguerre, line 63:
```
// detect the underlying architecture(amd64/arm64)
// we have to do this rather than just using runtime.GOARCH  because......
arch := utils.UnderlyingArch()
return []string{fmt.Sprintf(layerFmtGzip, imageType, runtime.GOOS, arch)} 
```

utils/arch.go
```diff
@@ -0,0 +1,31 @@
+package utils
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+// UnderlyingArch detects the underlying architecture(amd64/arm64) of the system
+func UnderlyingArch() string {
```
kaidaguerre, line 10:
```
// we need this because...
```

utils/arch.go
```diff
@@ -0,0 +1,31 @@
...
+	name := "uname"
+	arg0 := "-m"
+
+	cmd := exec.Command(name, arg0)
```
kaidaguerre, line 14: in this case (for a change :) ) I think this would be clearer

```
// shell out to uname command
cmd := exec.Command("uname", "-m)
```

utils/arch.go
```diff
@@ -0,0 +1,31 @@
...
+	switch underlyingArch {
+	// darwin and linux systems return "x86_64"
+	case "x86_64":
+		return "amd64"
+	// linux systems return "aarch64"
+	case "aarch64":
+		return "arm64"
+	default:
+		return underlyingArch
+	}
```
kaidaguerre, line 30: maybe return error if unexpected value?

kaidaguerre: @binaek thoughts?

utils/arch.go
binaek:
```
func UnderlyingArch() (string,error) {
...
}
```
Return the `error` if you face any

utils/arch.go
```diff
@@ -0,0 +1,31 @@
...
+
+// UnderlyingArch detects the underlying architecture(amd64/arm64) of the system
+func UnderlyingArch() string {
+	name := "uname"
+	arg0 := "-m"
+
+	cmd := exec.Command(name, arg0)
+	stdout, err := cmd.Output()
+	if err != nil {
+		fmt.Errorf("Error finding the underlying architecture: %s", err.Error())
```
binaek, line 17: @pskrbasu this creates an error object. doesn't actually print it. Return it. Look above

utils/arch.go
binaek: @kaidaguerre I agree

ociinstaller/mediatypes.go
```diff
@@ -60,7 +60,12 @@ func MediaTypeForPlatform(imageType ImageType) []string {
 	case ImageTypeDatabase:
 		return []string{fmt.Sprintf(layerFmtTar, imageType, runtime.GOOS, arch)}
 	case ImageTypeFdw:
-		return []string{fmt.Sprintf(layerFmtGzip, imageType, runtime.GOOS, utils.UnderlyingArch())} // Detect the underlying architecture(amd64/arm64)
+		// detect the underlying architecture(amd64/arm64)
+		// we have to do this rather than just using runtime.GOARCH, because runtime.GOARCH does not give us
+		// the actual underlying architecture of the system(GOARCH can be changed during runtime)
+		arch, err := utils.UnderlyingArch()
+		utils.FailOnErrorWithMessage(err, "failed to find underlying architecture")
```
kaidaguerre, line 67: bubble error up rather than failing - FailOnErrorWithMessage should only really be called from command package or where we are sure we need to just bomb out

kaidaguerre (changes requested):

ociinstaller/mediatypes.go
```diff
@@ -50,26 +51,33 @@ const (
...
-		return []string{fmt.Sprintf(layerFmtTar, imageType, runtime.GOOS, arch)}
+		return []string{fmt.Sprintf(layerFmtTar, imageType, runtime.GOOS, arch)}, nil
 	case ImageTypeFdw:
-		return []string{fmt.Sprintf(layerFmtGzip, imageType, runtime.GOOS, arch)}
+		// detect the underlying architecture(amd64/arm64)
+		// we have to do this rather than just using runtime.GOARCH, because runtime.GOARCH does not give us
+		// the actual underlying architecture of the system(GOARCH can be changed during runtime)
+		arch, err := utils.UnderlyingArch()
+		if err != nil {
+			return []string{}, err
```
kaidaguerre, line 68: nil, err

ociinstaller/mediatypes.go
```diff
@@ -50,26 +51,33 @@ const (
...
 		if runtime.GOOS == constants.OSDarwin && arch == constants.ArchARM64 {
 			// add the amd64 layer as well, so that we can fall back to it
 			// this is required for plugins which don't have an arm64 build yet
 			pluginMediaTypes = append(pluginMediaTypes, fmt.Sprintf(layerFmtGzip, imageType, constants.ArchAMD64, runtime.GOOS))
 		}
-		return pluginMediaTypes
+		return pluginMediaTypes, nil
 	}
-	return []string{}
+	return []string{}, nil
```
kaidaguerre, line 80: can we ever get here?

kaidaguerre: shouldn't we add a default to the switch statement which also returns an error? Invalid imagetype

kaidaguerre (changes requested):

ociinstaller/mediatypes.go
pskrbasu: will break for dashboard commands where imageType=assets

PR #1928, opened 2022-04-22
Prevents backup creation if there is an instance of 'pg12' running. Closes #1916

### LINKED ISSUE #1916: `pg14` installation should fail if there are instances of `pg12` already running

binaek:
If an instance of the old `pg12` service is still running from the `install-dir`, `steampipe` will error out and not create the backup (and also install the new `pg14`).

```
$ steampipe service start
Error: cannot start DB backup - an instance is still running. To stop running services, use steampipe service stop
```

### DIFF
--- a/db/db_local/backup.go
+++ b/db/db_local/backup.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 
 	"github.com/shirou/gopsutil/process"
 	"github.com/turbot/go-kit/helpers"
@@ -14,7 +15,11 @@ import (
 	"github.com/turbot/steampipe/filepaths"
 )
 
-type dbStartConfig struct {
+var (
+	ErrDbInstanceRunning = fmt.Errorf("cannot start DB backup - an instance is still running. To stop running services, use %s ", constants.Bold("steampipe service stop"))
+)
+
+type db12StartConfig struct {
 	cmd    *exec.Cmd
 	port   int
 	dbName string
@@ -30,6 +35,10 @@ func prepareBackup(ctx context.Context) (*string, error) {
 	if !needs {
 		return nil, nil
 	}
+	// short circuit if there is an instance running
+	if err := errIfInstanceRunning(ctx, location); err != nil {
+		return nil, err
+	}
 	config, err := startDatabaseInLocation(ctx, location)
 	if err != nil {
 		return nil, err
@@ -43,7 +52,24 @@ func prepareBackup(ctx context.Context) (*string, error) {
 	return &config.dbName, os.RemoveAll(location)
 }
 
-func takeBackup(ctx context.Context, config *dbStartConfig) error {
+func errIfInstanceRunning(ctx context.Context, location string) error {
+	processes, err := FindAllSteampipePostgresInstances(ctx)
+	if err != nil {
+		return err
+	}
+	for _, p := range processes {
+		cmdLine, err := p.CmdlineWithContext(ctx)
+		if err != nil {
+			continue
+		}
+		if strings.HasPrefix(cmdLine, filepaths.SteampipeDir) {
+			return ErrDbInstanceRunning
+		}
+	}
+	return nil
+}
+
+func takeBackup(ctx context.Context, config *db12StartConfig) error {
 	cmd := exec.CommandContext(
 		ctx,
 		getPgDumpBinaryExecutablePath(),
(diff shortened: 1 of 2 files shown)


db/db_local/backup.go
```diff
@@ -7,14 +7,19 @@ import (
...
 
 	"github.com/shirou/gopsutil/process"
 	"github.com/turbot/go-kit/helpers"
 	"github.com/turbot/steampipe/constants"
 	"github.com/turbot/steampipe/filepaths"
 )
 
-type dbStartConfig struct {
+var (
+	ErrDbInstanceRunning = fmt.Errorf("cannot start DB backup - an instance is still running. To stop running services, use %s ", constants.Bold("steampipe service stop"))
```
kaidaguerre, line 19: does this need to be exported?

db/db_local/locations.go
```diff
@@ -48,9 +48,9 @@ func getDataLocation() string {
 	return loc
 }
 
-// location where the dump file will be stored, so that it can be later restored after connections
+// tar file where the dump file will be stored, so that it can be later restored after connections
 // refresh in a new installation
-func getBackupLocation() string {
+func getBackupFile() string {
```
kaidaguerre, line 53: can we make all functions in this module consistent with the function names in the filepaths package, i.e. instead of 
```
getPostgresBinaryExecutablePath
```
it would be
```
postgresBinaryExecutableFilePath
```

and this one would be 
```
databaseBackupFilePath
```

kaidaguerre: also at the top of this file is 
```
const ServiceExecutableRelativeLocation = "/db/12.1.0/postgres/bin/postgres"
```

that is not right is it? Can we not autogenerate that path based on current PG version?

db/db_local/backup.go
```diff
@@ -7,14 +7,19 @@ import (
...
+)
+
+type db12StartConfig struct {
```
kaidaguerre, line 22: why db12StartConfig?

db/db_local/backup.go
```diff
@@ -30,6 +35,10 @@ func prepareBackup(ctx context.Context) (*string, error) {
 	if !needs {
 		return nil, nil
 	}
+	// short circuit if there is an instance running
```
kaidaguerre, line 38: // fail if there is a db instance running

kaidaguerre (changes requested):

db/db_local/locations.go
binaek: done

db/db_local/backup.go
binaek: Renamed to `pg12RunningInfo`

db/db_local/backup.go
binaek: done

db/db_local/install.go
```diff
@@ -70,23 +71,29 @@ func EnsureDBInstalled(ctx context.Context) (err error) {
...
-		return fmt.Errorf("Download & install steampipe-postgres-fdw... FAILED!")
-	}
-
 	statushooks.SetStatus(ctx, "Preparing backups...")
 	var dbName *string
 	// call prepareBackup to generate the db dump file if necessary
 	// NOTE: this returns the existing database name - we use this when creating the new database
 	if d, err := prepareBackup(ctx); err != nil {
+		if errors.Is(err, errDbInstanceRunning) {
+			// remove the installation, since otherwise, the backup won't get triggered,
```
kaidaguerre, line 80: could be on one line

db/db_local/backup.go
```diff
@@ -79,7 +105,7 @@ func takeBackup(ctx context.Context, config *dbStartConfig) error {
 }
 
 // startDatabaseInLocation starts up the postgres binary in a specific installation directory
```
kaidaguerre, line 107: does this always start PG12 instance? presumably as it returns `pg12RunningInfo` - in which case say so

db/db_local/backup.go
```diff
@@ -43,11 +52,28 @@ func prepareBackup(ctx context.Context) (*string, error) {
...
+			continue
+		}
+		if strings.HasPrefix(cmdLine, filepaths.SteampipeDir) {
+			return errDbInstanceRunning
+		}
+	}
+	return nil
+}
+
+func takeBackup(ctx context.Context, config *pg12RunningInfo) error {
```
kaidaguerre, line 72: // backup the pg12 instance public schema using pg_dump

kaidaguerre (changes requested):

db/db_local/install.go
binaek: 👍

PR #1979, opened 2022-04-29
Fixes issue where 'pg_dump' and 'pg_restore' depends on system libraries instead of packaged ones. Closes #1972

### LINKED ISSUE #1972: postgres binaries should use libraries from the installation and not the system
In some circumstances, `pg_dump` and `pg_restore` are using `libpq` that is installed in the system. This may fail, since `steampipe` may be installed in systems where there's not `postgres` pre-installed.

binaek:


db/db_local/backup.go
```diff
@@ -376,3 +380,55 @@ func getTableOfContentsFromBackup(ctx context.Context) ([]string, error) {
...
+		ctx,
+		pgRestoreBinaryExecutablePath(),
+		args...,
+	)
+	log.Println("[TRACE] pg_restore command:", cmd.String())
+	addLDPath(ctx, cmd)
+	return cmd
+}
+
+func addLDPath(ctx context.Context, cmd *exec.Cmd) {
```
kaidaguerre, line 406: add comment - why are we doing this
Change name to UpdateDynamicLibPath

kaidaguerre (changes requested):

db/db_local/backup.go
binaek: done

PR #2131, opened 2022-05-30
Improve the response to "steampipe plugin update all". Closes #2125

### LINKED ISSUE #2125: improve the response to "steampipe plugin install all"
We say:

```
$ steampipe plugin update all

Skipped the following plugin:

Plugin:   all@latest
Reason:   Not installed

To install plugin which is not installed, please run steampipe plugin install all
```

Proposed:

```
$ steampipe plugin update all

Did you mean --all?
```

pskrbasu:
Screenshot 2022-05-30 at 5 45 26 PM

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -306,6 +306,14 @@ func runPluginUpdateCmd(cmd *cobra.Command, args []string) {
 		return
 	}
 
+	if len(plugins) > 0 && !(cmdconfig.Viper().GetBool("all")) && plugins[0] == "all" {
+		// improve the response to wrong argument "steampipe plugin update all"
+		fmt.Println()
+		utils.ShowError(ctx, fmt.Errorf("Did you mean %s?", constants.Bold("--all")))
+		fmt.Println()
+		return
+	}
+
 	state, err := statefile.LoadState()
 	if err != nil {
 		utils.ShowError(ctx, fmt.Errorf("could not load state"))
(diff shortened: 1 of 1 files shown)


judell:
Thanks @pskrbasu!

cmd/plugin.go
```diff
@@ -508,6 +516,11 @@ func resolveUpdatePluginsFromArgs(args []string) ([]string, error) {
 		// we can't allow update and install at the same time
 		return nil, fmt.Errorf("%s cannot be used when updating specific plugins", constants.Bold("`--all`"))
 	}
+
+	// if len(plugins) > 0 && !(cmdconfig.Viper().GetBool("all")) && plugins[0] == "all" {
```
kaidaguerre, line 520: remove comment

cmd/plugin.go
```diff
@@ -306,6 +306,14 @@ func runPluginUpdateCmd(cmd *cobra.Command, args []string) {
 		return
 	}
 
+	if len(plugins) > 0 && !(cmdconfig.Viper().GetBool("all")) && plugins[0] == "all" {
```
kaidaguerre, line 309: why do we need `!(cmdconfig.Viper().GetBool("all"))` in the if?

kaidaguerre (changes requested):
I wonder if we should only trigger this logic if plugin is not found, i.e. if there is an `all` plugin installed, we update it?

cmd/plugin.go
pskrbasu: to check if there's any `--all` flag

PR #2134, opened 2022-05-31
Add support for --dashboard flag to service restart. Closes #2106

### LINKED ISSUE #2106: Add support for `--dashboard` flag to `service restart`

pskrbasu:


cmd/service.go
```diff
@@ -136,8 +136,11 @@ func serviceRestartCmd() *cobra.Command {
 	cmdconfig.
 		OnCmd(cmd).
 		AddBoolFlag(constants.ArgHelp, "h", false, "Help for service restart").
-		AddBoolFlag(constants.ArgForce, "", false, "Forces the service to restart, releasing all open connections and ports")
-
+		AddBoolFlag(constants.ArgForce, "", false, "Forces the service to restart, releasing all open connections and ports").
+		// dashboard server
+		AddBoolFlag(constants.ArgDashboard, "", false, "Run the dashboard webserver with the service").
+		AddStringFlag(constants.ArgDashboardListen, "", string(dashboardserver.ListenTypeNetwork), "Accept connections from: local (localhost only) or network (open) (dashboard)").
```
kaidaguerre, line 142: I don't think `ArgDashboardPort` and `ArgDashboardListen` should be exposed as args - we should just reuse values for the currently running dashboard service (are these in the dashboard state?)

kaidaguerre (changes requested):

cmd/service.go
pskrbasu: makes sense. fixed

pskrbasu:
Closing this, since the current behaviour seems appropriate.
`service restart` restarts the dashboard server, if it was running previously using the same `listen` and `ports`

PR #2181, opened 2022-06-17
Fix issue where check/templates were not getting updated even when the template file has been updated. Closes #2180

### LINKED ISSUE #2180: `check/templates` not getting updated even when a template has been updated
The `$HOME/.steampipe/check/templates` store the check templates, now when a template is updated from code and built, the changes do not get reflected in the `$HOME/.steampipe/check/templates/<template-name>/output.tmpl` file, which causes the output to be generated by the old template.

pskrbasu:


control/controldisplay/formatter_fs.go
```diff
@@ -13,30 +13,24 @@ import (
 //go:embed templates/*
 var builtinTemplateFS embed.FS
 
-// EnsureTemplates scans the '$STEAMPIPE_INSTALL_DIR/templates' directory and
-// copies over any missing templates as defined in the 'templates' package
+// WriteTemplates scans the '$STEAMPIPE_INSTALL_DIR/templates' directory and
+// copies over all the templates defined in the 'templates' package
```
kaidaguerre, line 17: add a comment to explain why we always write

kaidaguerre (changes requested):

Issue #2184, opened 2022-06-17
Cannot provide query parameters with `.` in steampipe terminal.


nmische:
**Describe the bug**

When attempting to create a mod query with a param that expects a `.` in the value, the steampipe terminal retruns:

```
Error: syntax error at or near "query" (SQLSTATE 42601)
```

**Steampipe version (`steampipe -v`)**

steampipe version 0.14.6

**To reproduce**

Create a simple mod with the following query:

```
query "test" {
  sql = <<-EOT
    select
     'foo' as foo
    where
     'foo.bar' = $1
  EOT
  param "test_input" {
    description = "test input"
  }
}
```

Run the query from the steampipe terminal, attempting to pass in a value of `foo.bar`:

```
> query.test(test_input => "foo.bar")
Error: syntax error at or near "query" (SQLSTATE 42601)
```

**Expected behavior**
Expect to be able to provide parameter values with a `.` character.

**Additional context**
Paramater values without a `.` do not cause this error:

```
> query.test(test_input => "foo")
+-----+
| foo |
+-----+
+-----+
```

PR #2185, opened 2022-06-17
Strip params before calling ParseResourceName

nmische:
Addresses #2184.

### DIFF
--- a/workspace/workspace_queries.go
+++ b/workspace/workspace_queries.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"strings"
 
 	typehelpers "github.com/turbot/go-kit/types"
 	"github.com/turbot/steampipe/steampipeconfig/modconfig"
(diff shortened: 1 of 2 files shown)


workspace/workspace_queries.go
```diff
@@ -182,6 +183,10 @@ func (w *Workspace) getQueryFromFile(filename string) (string, bool, error) {
 
 // does this resource name look like a control or query
 func isNamedQueryOrControl(name string) bool {
+	openBracketIdx := strings.Index(name, "(")
```
nmische, line 186: You could probably make a case for adding this logic in [`ParseResourceName`](https://github.com/turbot/steampipe/blob/main/steampipeconfig/modconfig/parsed_name.go#L14-L39), but I opted to make a more localized fixed given how broadly `ParseResourceName` is used throughout the codebase.

kaidaguerre: agreed, that makes sense for now

kaidaguerre: I would either add a short comment here explaining why it it necessary

kaidaguerre (changes requested):
Many thanks for this, looks good.

Added one minor change request

nmische:
@kaidaguerre: I believe I have addressed your feedback. Let me know if you need anything else for this one. Thanks!

PR #2339, opened 2022-08-18
Provide feedback in interactive mode when loading a workspace with a mod.sp, and loading all files recursively. Closes #2323

### LINKED ISSUE #2323: Provide feedback in interactive mode when  loading a workspace with a mod.sp, and loading all files recursively
When running Steampipe, it searches the current folder and the _up parent folder tree_ to find a mod.sp file containing a mod definition.  So if you are running Steampipe from within a child folder of a mod folder, it recognises that mod as the current workspace mod.

If a mod.sp is found, a recursive search is executed to find any .sp underneath the mod folder directory. These are parsed and loaded into the workspace.
Any .sql files found recursively under the mod directory are converted into query resources in the workspace

Currently no indication is given that this is happening, which can be confusing, particularly if you are inadvertently running in the child folder of a mod.

Add feedback when starting up:

```
Modfile found at path xxx/xxx/xxx/mod.sp`
Loading .sp and .sql files recursively underneath xxx/xxx/xxx
Found 100 files
```

pskrbasu:


kaidaguerre:
more complex than initially thought - closing for now

PR #2343, opened 2022-08-19
Add better help messages for mod plugin requirements not satisfied error

pskrbasu:
- install/update
Screenshot 2022-08-23 at 5 47 46 PM

- only install
Screenshot 2022-08-23 at 5 48 00 PM

- only update
Screenshot 2022-08-23 at 5 48 18 PM

### DIFF
--- a/pkg/workspace/workspace_require.go
+++ b/pkg/workspace/workspace_require.go
@@ -149,14 +149,22 @@ func pluginVersionError(pluginsNotInstalled []requiredPluginVersion) string {
 	}
 
 	// add help message for missing plugins
-	msg := fmt.Sprintf("\nPlease install the %s with: \n", utils.Pluralize("plugin", len(pluginsNotInstalled)))
+	msg := fmt.Sprintf("\nPlease install/update the %s with: \n", utils.Pluralize("plugin", len(pluginsNotInstalled)))
 	notificationLines = append(notificationLines, msg)
 
-	for _, req := range pluginsNotInstalled {
+	for i, req := range pluginsNotInstalled {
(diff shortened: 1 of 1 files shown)


kaidaguerre (changes requested):
Looks good, one thing - would be nice to be smart about `install/update`

i.e. if there are only installations, use `install`, if there are only updates use `update`, if there are both use `install/update`

The challenge is to do this in a non-messy way :)

pkg/workspace/workspace_require.go
```diff
@@ -148,8 +148,46 @@ func pluginVersionError(pluginsNotInstalled []requiredPluginVersion) string {
...
+		}
+	}
+
 	// add blank line (hack - bold the empty string to force it to print blank line as part of error)
 	notificationLines = append(notificationLines, fmt.Sprintf("%s", constants.Bold("")))
 
 	return strings.Join(notificationLines, "\n")
 }
+
+// function to check whether the missing plugins require to be installed or updated, or both
```
kaidaguerre, line 176: could be tweaked to 
```
// function to check whether the missing plugins require to be installed or updated, or both
func checkInstallOrUpdate(pluginsNotInstalled []requiredPluginVersion) string {
	var updateFlag, installFlag bool
	
	for _, req := range pluginsNotInstalled {
		if strings.Contains(req.installedVersion, "none") {
			installFlag = true
		} else {
			updateFlag = true
		}
	}
	
	if updateFlag {
		if installFlag {
			return "install/update"
		} else {
			return "update"
		}
	}
	return "install"
}
```

kaidaguerre (changes requested):

PR #2405, opened 2022-09-01
Fix initdb failures due to invalid locale. Closes #2368

### LINKED ISSUE #2368: `initdb` failure due to invalid locale
Currently, we have a check for the `invalid locale` settings issue
https://github.com/turbot/steampipe/blob/305fb9107b75d6756b66e69cbe55d290734ae36d/main.go#L91-L118

However, it seems like sometimes the invalid locale is not the sole issue for the initdb failure. 
https://steampipe.slack.com/archives/C01UECB59A7/p1660657259557199

We need to figure out a way to expand the checks to cover this case.

pskrbasu:


pkg/db/db_local/install.go
```diff
@@ -434,6 +434,16 @@ func initDatabase() error {
 	utils.LogTime("db_local.install.initDatabase start")
 	defer utils.LogTime("db_local.install.initDatabase end")
 
+	// initdb sometimes fail due to invalid locale settings, to avoid this we update
+	// the locale settings to use 'C' only for the initdb process to complete, and
+	// then return back to the existing locale settings of the user.
```
kaidaguerre, line 439: you need to read the previous value of LC_ALL and reset it - only unset if there is no original value

kaidaguerre (changes requested):

PR #2495, opened 2022-10-09
Update CONTRIBUTING.md

akashgreninja:
added some lines and removed because ,does not make sense for it to be there


kaidaguerre (commented):
Thanks for the PR. I'm going to reject it however.

While I agree the phrasing is not the most elegant, it is actually more grammatically correct before the change. The word 'because' does serve  a function in the sentence.

PR #2523, opened 2022-10-13
Deprecate `workspace-chdir`, replace with `mod-location`

### LINKED ISSUE #2511: Deprecate workspace-chdir, replace with mod-location

michizhou:
Deprecates the existing `workspace-chdir` argument and replaces it with new argument `mod-location`. Closes #2511.


cmd/root.go
```diff
@@ -83,15 +83,15 @@ func InitCmd() {
 	defer utils.LogTime("cmd.root.InitCmd end")
 
 	rootCmd.PersistentFlags().String(constants.ArgInstallDir, filepaths.DefaultInstallDir, fmt.Sprintf("Path to the Config Directory (defaults to %s)", filepaths.DefaultInstallDir))
-	rootCmd.PersistentFlags().String(constants.ArgWorkspaceChDir, "", "Path to the workspace working directory")
+	rootCmd.PersistentFlags().String(constants.ArgModLocation, "", "Path to the workspace working directory")
```
kaidaguerre, line 86: we need to deprecate the existing flag, rather than remove it altogether

desired behaviour is that if `--workspace-chdir` is used a deprecation warning is shown but Steampipe respects the arg value

so we need to add and bind both the old and new flags, but deprecate the old one with code like this: 

```
workspaceChDirFlag := rootCmd.PersistentFlags().Lookup(constants.ArgWorkspaceChDir)
	workspaceChDirFlag.Deprecated = "use --mod-location"

	err= viper.BindPFlag(constants.ArgWorkspaceChDir, workspaceChDirFlag))`
```

we also need to copy the value of `workspace-chdir` into `mod-location`. I suggest adding a function like this: 

```
func handleArgDeprecations() {
	if viper.GetString(constants.ArgModLocation) == "" {
		viper.Set(constants.ArgModLocation, viper.GetString(constants.ArgWorkspaceChDir))
	}
}
```

and calling it from `PersistentPreRun` https://github.com/turbot/steampipe/blob/1c2d62ea56c87c4fe54283c75c69c25145b055b1/cmd/root.go#L42

kaidaguerre (changes requested):
Thanks for the PR!

We need to ensure `--workspace-chdir` continues to work as well for now. See comment below.

kaidaguerre:
@michizhou please could you rebase this on main, then we can get it merged

michizhou:
@kaidaguerre I have rebased the branch on the latest state of main. Please let me know if additional changes are needed. Otherwise, please feel free to merge this at your discretion.

PR #2548, opened 2022-10-18
Fixes issue where prompt fails to show up when service installation runs in interactive mode

binaek:


binaek:
https://user-images.githubusercontent.com/1114038/196655866-da9ba5ae-118f-4119-afc9-e40cfd0e1ef1.mov

pkg/interactive/interactive_client_init.go
```diff
@@ -39,6 +39,8 @@ func (c *InteractiveClient) handleInitResult(ctx context.Context, initResult *db
 		return
 	}
 
+	initResult.AddWarnings("this is a hard coded watning")
```
kaidaguerre, line 42: please remove test code

kaidaguerre (changes requested):

pkg/interactive/interactive_client_init.go
binaek: Oops!

PR #2556, opened 2022-10-19
Fixes issue where acceptance tests are not running locally.

binaek:


tests/acceptance/test_files/mod.bats
```diff
@@ -3,525 +3,525 @@ load "$LIB_BATS_SUPPORT/load.bash"
 
 ############### QUERIES ###############
 
-@test "query with default params and no params passed through CLI" {
-  cd $FUNCTIONALITY_TEST_MOD
-  run steampipe query query.query_params_with_all_defaults --output json
+# @test "query with default params and no params passed through CLI" {
```
kaidaguerre, line 6: why is this all commented out? do we still need it?

kaidaguerre (changes requested):

PR #2567, opened 2022-10-20
Fixes issue where the CLI would fail to connect to local service if there are credential files in `~/.postgresql` 

binaek:


pkg/db/db_local/ssl.go
```diff
@@ -251,19 +201,31 @@ func generateServerCertificates(caCertificateData *x509.Certificate, caPrivateKe
...
 func sslStatus() string {
-	status := sslMode()
-	if status == "require" {
+	if serverCertificateAndKeyExist() {
 		return "on"
 	}
 	return "off"
 }
 
 // derive ssl mode from the prsesnce of the server certificate and key file
```
kaidaguerre, line 210: typo in comment

pkg/db/db_common/config.go
```diff
@@ -0,0 +1,15 @@
+package db_common
```
kaidaguerre, line 1: `config` seems a bit of a generic filename - maybe `ssl` / `cert`?

kaidaguerre (changes requested):

Issue #2560, opened 2022-10-20
Add terminal option to disable auto-complete


e-gineer:
For very large environments autocomplete in the interactive terminal can be annoying slow for typing. While it should be faster, it would also be helpful at times to just disable autocomplete.

I'd propose a setting for terminal - https://steampipe.io/docs/reference/config-files/terminal

```
auto_complete = false
```

TomerPacific:
Hey @e-gineer , I'd like to help out here. Any details I should be aware of before starting?

e-gineer:
We'd love your help @TomerPacific! As a starting point, I imagine the implementation would be similar to `multi`

https://github.com/turbot/steampipe/search?l=Go&q=multi

TomerPacific:
@e-gineer , I have tried looking for a way to set up a local development environment, but couldn't find anything.
Is there a resource you can point me to?

I saw the [Contributing ](https://github.com/turbot/steampipe/blob/main/CONTRIBUTING.md)guide, but it did not have this information. Especially if this feature is related only to the terminal.

I want to have a way to test out my changes.

TomerPacific:
@e-gineer - seems like I have been bamboozled by [Mr-Destructive](https://github.com/Mr-Destructive) (what an appropriate name).

PR #2584, opened 2022-10-24
add auto-completion toggle option

Mr-Destructive:
I've added the config for setting the auto-completion toggle, but it isn't able to switch off the auto-completion.
Can someone please guide me in the appropriate direction?
addressing #2560

### DIFF
--- a/pkg/interactive/interactive_client.go
+++ b/pkg/interactive/interactive_client.go
@@ -580,20 +580,20 @@ func (c *InteractiveClient) shouldExecute(line string, namedQuery bool) bool {
 }
 
 func (c *InteractiveClient) queryCompleter(d prompt.Document) []prompt.Suggest {
-	if !c.isInitialised() {
-		return nil
-	}
 	if !cmdconfig.Viper().GetBool(constants.ArgAutoComplete) {
 		return nil
 	}
-
-	var s []prompt.Suggest
+	if !c.isInitialised() {
+		return nil
+	}
 
 	text := strings.TrimLeft(strings.ToLower(d.Text), " ")
 
 	if len(c.interactiveBuffer) > 0 {
 		text = strings.Join(append(c.interactiveBuffer, text), " ")
 	}
+
+	var s []prompt.Suggest
 
 	if len(d.CurrentLine()) == 0 && !c.autocompleteOnEmpty {
 		// if nothing has been typed yet, no point
--- a/pkg/interactive/interactive_client.go
+++ b/pkg/interactive/interactive_client.go
@@ -347,10 +347,6 @@ func (c *InteractiveClient) runInteractivePrompt(ctx context.Context) (ret utils
 		}),
 		// ignore suppressed ASCII codes
 		prompt.OptionAddASCIICodeBind(c.suppressOnInput()...),
-		prompt.OptionAddKeyBind(prompt.KeyBind{
-			Key: prompt.ControlC,
-			Fn:  func(b *prompt.Buffer) { c.enabeAutoComplete(b) },
-		}),
 	)
 	// set this to a default
 	c.autocompleteOnEmpty = false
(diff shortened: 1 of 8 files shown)


pkg/interactive/interactive_client.go
```diff
@@ -347,6 +347,10 @@ func (c *InteractiveClient) runInteractivePrompt(ctx context.Context) (ret utils
 		}),
 		// ignore suppressed ASCII codes
 		prompt.OptionAddASCIICodeBind(c.suppressOnInput()...),
+		prompt.OptionAddKeyBind(prompt.KeyBind{
```
binaek, line 350: Let's not add this. This may interfere with the other handler for `Ctrl+C` which is responsible for breaking out of a multiline session. Moreover I don't think we want to toggle auto complete based on Ctrl+C. `Ctrl+C` is more of a `cancel current action` than `disable/enable this feature`

pkg/interactive/interactive_client.go
```diff
@@ -571,6 +575,10 @@ func (c *InteractiveClient) shouldExecute(line string, namedQuery bool) bool {
 		// statement has terminating ';'
 		return true
 	}
+	if !cmdconfig.Viper().GetBool(constants.ArgAutoComplete) {
```
binaek, line 578: The `shouldExecute` function is called from the `executor` which is called only when the query needs execution. `Execution` happens later. 

I think a variant of this should go in at `InteractiveClient.queryCompleter`.

binaek (changes requested):
@Mr-Destructive Thank you for raising this and for the help! I added a few comments which should help to get this working. Can you take a look?

Mr-Destructive:
Yup! Thank you that is now working by setting `.auto_complete on` or `off`.

pkg/interactive/interactive_client.go
```diff
@@ -587,15 +583,18 @@ func (c *InteractiveClient) queryCompleter(d prompt.Document) []prompt.Suggest {
 	if !c.isInitialised() {
 		return nil
 	}
+	if !cmdconfig.Viper().GetBool(constants.ArgAutoComplete) {
```
binaek, line 586: This should probably be the first thing we do. Is there any reason to check for `c.isInitialised()` if `auto complete` is disabled anyway?

pkg/interactive/interactive_client.go
```diff
@@ -587,15 +583,18 @@ func (c *InteractiveClient) queryCompleter(d prompt.Document) []prompt.Suggest {
...
+		return nil
+	}
+
+	var s []prompt.Suggest
```
binaek, line 590: Do we need this?

binaek (changes requested):
@Mr-Destructive thank you for the updates. I requested a couple of small changes.

pkg/interactive/interactive_client.go
Mr-Destructive: Are you talking about s? ya it will be used in the return value

pkg/interactive/interactive_client.go
binaek: Does this need to be there? The `shouldExecute` does not get called till the actual Execution time.

Autocomplete setting should not influence the decision to execute. Can you please remove this?

pkg/interactive/interactive_client.go
```diff
@@ -669,3 +677,7 @@ func (c *InteractiveClient) startCancelHandler() chan bool {
 	}()
 	return quitChannel
 }
+
+func (c *InteractiveClient) enabeAutoComplete(buffer *prompt.Buffer) {
```
binaek, line 681: It doesn't look like this function is getting called from anywhere. Does this still need to be here? You may remove it if not!

binaek (changes requested):
Hey @Mr-Destructive . Thank you for the revisions. I requested a couple of changes (mostly around cleanup). Can you please take a look?

PR #2626, opened 2022-10-29
Added contributors list

atulyakhatri:
Updated contributors list in the documentation !!


kaidaguerre:
@atulyakhatri many thanks for the PR, however we'd rather not include this widget.

PR #2720, opened 2022-11-09
Fixes issue where interactive prompt was not showing timing data for 'json', 'csv' and 'line' outputs.

binaek:


pkg/query/queryexecute/execute.go
```diff
@@ -27,7 +27,7 @@ func RunInteractiveSession(ctx context.Context, initData *query.InitData) {
 
 	// print the data as it comes
 	for r := range resultsStreamer.Results {
-		display.ShowOutput(ctx, r)
+		display.ShowOutput(ctx, r, display.ShowTimingIfEnabled())
```
kaidaguerre, line 30: ShowTimingIfEnabled is a bit counterintuittive - why do we need an option for this - can we not just.... show it if it;s enabled

It would feel more intuitive to have a `diusable` timing option

kaidaguerre (changes requested):

pkg/query/queryexecute/execute.go
binaek: `ShowTimingIfEnabled` peeks into the `viper` config and only displays the timing data if it is enabled in `viper`

This is meant to cover `.timing` commands in the interactive prompt

pkg/display/display.go
```diff
@@ -25,8 +25,57 @@ import (
 	"golang.org/x/text/message"
 )
 
+type ShowOutputConfiguration struct {
```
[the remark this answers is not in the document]
binaek: resolved

pkg/display/display.go
```diff
@@ -25,8 +25,49 @@ import (
...
+type DisplayConfiguration struct {
+	timing *bool
+}
+
+type DisplayOption = func(config *DisplayConfiguration)
+
+// ShowTimingOnOutput only enables timing if the current output mode is the one provided
+func ShowTimingOnOutput(output string) DisplayOption {
+	return func(o *DisplayConfiguration) {
+		if o.timing == nil {
```
kaidaguerre, line 37:
```
                if o.timing == nil {
			EnableTiming()(o)
		}
```
what is this doing?

kaidaguerre: is it needed?

kaidaguerre (changes requested):

PR #2830, opened 2022-12-02
Adds consistent error handling and wrapping across steampipe cli

binaek:
Based from the HEAD of #3032


binaek:
Screenshot 2023-01-02 at 8 52 19 PM

kaidaguerre:
```Create error with message and detail
func validateData(data int) error {
  if data > 10 {
    return sperr.Wrap(
      sperr.New("invalid argument: %d", data),
      sperr.WithDetail("error occurred with %d argument", data),
    )
  }
  return nil
}```

wouldn't this be 
```
Create error with message and detail
func validateData(data int) error {
  if data > 10 {
    return sperr.New("invalid argument: %d", data).WithDetail("error occurred with %d argument", data),
    
  }
  return nil
}
```

binaek:
There's too much change which are inconsistent with how things work.

This needs to be done - as an ongoing process and should not be attempted as one big blow.

An ongoing approach is necessary this changes will touch pretty much all layers of the codebase and as such needs to be done carefully

PR #2835, opened 2022-12-06
Fixes issue where implicit services are not shutting down when the last instance of steampipe exits

binaek:


cmd/service.go
```diff
@@ -317,7 +317,7 @@ func runServiceInForeground(ctx context.Context, invoker constants.Invoker) {
 			dashboardserver.StopDashboardService(ctx)
 			// if we have received this signal, then the user probably wants to shut down
 			// everything. Shutdowns MUST NOT happen in cancellable contexts
-			count, err := db_local.GetCountOfThirdPartyClients(context.Background())
+			_, count, err := db_local.GetClientCount(context.Background())
```
kaidaguerre, line 320: maybe return a struct?
```
{
     steampipeClientCount int
     totalCount int
}
```

kaidaguerre: or something

pkg/db/db_local/install.go
```diff
@@ -40,7 +40,7 @@ If you need to restore the contents of your public schema, please open an issue
 
 // EnsureDBInstalled makes sure that the embedded pg database is installed and running
 func EnsureDBInstalled(ctx context.Context) (err error) {
-	utils.LogTime("db_local.EnsureDBInstalled start")
+	utils.LogTime("db_local.EnsureDBInstallation start")
```
kaidaguerre, line 43: EnsureDBInstallation  ???
Function is still ``EnsureDBInstalled`

kaidaguerre (changes requested):

pkg/db/db_local/install.go
binaek: done

cmd/service.go
binaek: done

PR #2837, opened 2022-12-06
Adds tests around service lifecycle

binaek:


e-gineer:
The notes directory seems new with markdown files in it ... Do we want to have docs in a place like this?

CC @kaidaguerre @johnsmyth

binaek:
> The notes directory seems new with markdown files in it ... Do we want to have docs in a place like this?
> 
> CC @kaidaguerre @johnsmyth

We wanted to start documenting some of the subtleties and complexities of the underlying postgres service and it's lifecycle.

Created the `notes` directory as a starting point. Maybe we should move the content to a Wiki for the steampipe repository?

PR #2846, opened 2022-12-07
Fixes issue where task running goroutine would try to access viper config as it is being written to

binaek:


cmd/root.go
```diff
@@ -66,12 +66,23 @@ var rootCmd = &cobra.Command{
 
 		createLogger()
 
+		// set up the global viper config with values from
```
kaidaguerre, line 69: `set up the global viper config with default values from ...`

pkg/task/config.go
```diff
@@ -0,0 +1,19 @@
...
+	runUpdateCheck bool
+}
+
+func newRunConfig() *runConfig {
+	return &runConfig{
+		runUpdateCheck: true,
+	}
+}
+
+type TaskRunConfig func(o *runConfig)
```
kaidaguerre, line 13: `RunOption`

kaidaguerre: I think our pattern is always - use an `Option` to set a `Configg`

cmd/root.go
binaek: done

pkg/task/config.go
binaek: done

PR #2861, opened 2022-12-12
Changes the serialization of NULL values in a CSV

### LINKED ISSUE #2847: CSV output for null values
**Describe the bug**
When asking for CSV output (using `--output csv`), NULL values as serialized as `<null>`.

steampipe version 0.17.4

**To reproduce**
```
$ steampipe query "select 1 as foo, NULL as null_value, 2 as bar" --output csv
foo,null_value,bar
1,<null>,2
```

While json output works as expected:
```
steampipe query "select 1 as foo, NULL as null_value, 2 as bar" --output json
[
 {
  "bar": 2,
  "foo": 1,
  "null_value": null
 }
]
```

**Expected behavior**
`NULL` values should be output with as `` (zero-length string).

The above example should be:
```
$ steampipe query "select 1 as foo, NULL as null_value, 2 as bar" --output csv
foo,null_value,bar
1,,2
```

**Additional context**
N/A

xcambar:
Fixes #2847

As discussed in the related issue, this PR offers the simplest change so that the CSV output of a query matches `psql`.
It fixes the general situation of NULL values, which used to be serialized as `<null>` and, with this PR, are now `` (empty string).

### DIFF
--- a/pkg/display/column.go
+++ b/pkg/display/column.go
@@ -84,3 +84,11 @@ func ParseJSONOutputColumnValue(val interface{}, col *queryresult.ColumnDef) (in
 		return ColumnValueAsString(val, col)
 	}
 }
+
+func ParseCSVOutputColumnValue(val interface{}, col *queryresult.ColumnDef) (string, error) {
+	if val == nil {
+		return "", nil
+	}
+
+	return ColumnValueAsString(val, col)
+}
--- a/pkg/display/display.go
+++ b/pkg/display/display.go
@@ -278,7 +278,12 @@ func displayCSV(ctx context.Context, result *queryresult.Result) {
 	// print the data as it comes
 	// define function display each csv row
 	rowFunc := func(row []interface{}, result *queryresult.Result) {
-		rowAsString, _ := ColumnValuesAsString(row, result.Cols)
+		columns := result.Cols
+		rowAsString := make([]string, len(columns))
+		for idx, val := range row {
+			val, _ := ParseCSVOutputColumnValue(val, columns[idx])
+			rowAsString[idx] = val
+		}
 		_ = csvWriter.Write(rowAsString)
 	}
 


binaek:
Hi @xcambar .

Thank you for the PR.

One question that comes to mind is how does a user differentiate between a `null` value in a cell versus an actual `empty string` value?

Although the same could be said for `<null>` - this was chosen with the assumption that the value `<null>` is probably not very common.

xcambar:
I want to highlight the following comment from the source code of the go's `eencoding/csv` package (that is used to format to csv in steampipe):

> // We used to quote empty strings, but we do not anymore (as of Go 1.4).
> // The two representations should be equivalent, but Postgres distinguishes
> // quoted vs non-quoted empty string during database imports, and it has
> // an option to force the quoted behavior for non-quoted CSV but it has
> // no option to force the non-quoted behavior for quoted CSV, making
> // CSV with quoted empty strings strictly less useful.
> // Not quoting the empty string also makes this package match the behavior
> // of Microsoft Excel and Google Drive.

https://github.com/golang/go/blob/master/src/encoding/csv/writer.go#L148-L155

This is a strong hint that NULL and empty string should not be differentiated in CSV, as go and `psql` are already aligned on that topic. It also means less code in steampipe. And less code is always better :p

binaek:
@xcambar thank you for the changes. While testing this, we noticed that this the updated behaviour has spilled over to the  `table` output to have `empty string` for `null` cells as well.

Can you update the PR to prevent that?

Let me know if you need someone to point out where to make the changes.

binaek:
@xcambar were you able to take a look into this?

xcambar:
I'm returning from holidays. I can have a look into it in the coming days but feel free to share some pointers to put me on the right track.

xcambar:
I've updated the code so that the changes only reflect in CSV output

pkg/display/column.go
```diff
@@ -84,3 +84,11 @@ func ParseJSONOutputColumnValue(val interface{}, col *queryresult.ColumnDef) (in
 		return ColumnValueAsString(val, col)
 	}
 }
+
+func ParseCSVOutputColumnValue(val interface{}, col *queryresult.ColumnDef) (string, error) {
```
binaek, line 88: Is it necessary to have a separate function for this?

Maybe we could pass in an option to `ColumnValueAsString` and achieve the same thing?

pkg/display/display.go
```diff
@@ -278,7 +278,12 @@ func displayCSV(ctx context.Context, result *queryresult.Result) {
 	// print the data as it comes
 	// define function display each csv row
 	rowFunc := func(row []interface{}, result *queryresult.Result) {
-		rowAsString, _ := ColumnValuesAsString(row, result.Cols)
+		columns := result.Cols
+		rowAsString := make([]string, len(columns))
```
binaek, line 282: `rowAsString := ColumnValueAsString(val, col, WithNullString(""))`

where we change `ColumnValueAsString` to accept a variadic set of options.

pkg/display/display.go
```diff
@@ -278,7 +278,12 @@ func displayCSV(ctx context.Context, result *queryresult.Result) {
...
+		for idx, val := range row {
+			val, _ := ParseCSVOutputColumnValue(val, columns[idx])
+			rowAsString[idx] = val
+		}
```
binaek, line 286: We shouldn't need this loop.

binaek (changes requested):
Hi @xcambar . Thank you for making the changes. 

I put in a couple of change requests. Please take a look when you get the chance.

I feel that by using a variadic configuration options in the `ColumnValuesAsString` function will help achieve the same thing and also open the door to future customisations when required.

I found this doc helpful: https://medium.com/soon-london/variadic-configuration-functions-in-go-8cef1c97ce99 
https://medium.com/soon-london/variadic-configuration-functions-in-go-8cef1c97ce99

xcambar:
A quick update: Feel free to close this PR, I will not apply the required changes. I can agree my PR is not ideal and I would understand it does not fit your requirements but my availability cannot afford the efforts.

I'd still suggest that Steampipe finds a way to follow psql's standards and outputs `null`s as empty strings, as discussed above.

Best,

binaek:
@xcambar Thank you so much for your efforts into bringing this to our notice and also for carrying out the initial work on this.

Let me take over from here and apply the changes.

binaek:
Implemented with #3019

binaek:
@xcambar I have implemented the `empty string for null` in the `2847-csv-output-for-null-values` branch.

When you get the chance, can you please give it a go and see if it fits with your expectations?

xcambar:
Thank you very much for picking up the work. Additionally, reading your PR taught me some more Go.

PR #2987, opened 2023-01-12
Fix issue where running query inside a mod directory without having installed its associated plugin first returns an error with 0 exit code. Closes #2986

### LINKED ISSUE #2986: Running query inside a mod directory without having installed its associated plugin first returns an error with 0 exit code
When I try to run steampipe query inside a mod but without having installed its associated plugin first, I see an error msg. But the exit code is 0.

pskrbasu:
Screenshot 2023-01-12 at 7 31 42 PM

Screenshot 2023-01-12 at 3 41 12 PM

Screenshot 2023-01-12 at 3 42 14 PM

### DIFF
--- a/pkg/query/queryexecute/execute.go
+++ b/pkg/query/queryexecute/execute.go
@@ -42,7 +42,7 @@ func RunBatchSession(ctx context.Context, initData *query.InitData) (int, error)
 	// wait for init
 	<-initData.Loaded
 	if err := initData.Result.Error; err != nil {
-		return -1, err
+		return 0, err
 	}
 
 	// display any initialisation messages/warnings
--- a/pkg/query/queryexecute/execute.go
+++ b/pkg/query/queryexecute/execute.go
@@ -34,7 +34,7 @@ func RunInteractiveSession(ctx context.Context, initData *query.InitData) {
 	}
 }
 
-func RunBatchSession(ctx context.Context, initData *query.InitData) int {
+func RunBatchSession(ctx context.Context, initData *query.InitData) (int, error) {
 	// start cancel handler to intercept interrupts and cancel the context
 	// NOTE: use the initData Cancel function to ensure any initialisation is cancelled if needed
 	contexthelpers.StartCancelHandler(initData.Cancel)
@@ -42,7 +42,7 @@ func RunBatchSession(ctx context.Context, initData *query.InitData) int {
 	// wait for init
 	<-initData.Loaded
 	if err := initData.Result.Error; err != nil {
-		error_helpers.FailOnError(err)
+		return -1, err
 	}
 
 	// display any initialisation messages/warnings
@@ -54,7 +54,7 @@ func RunBatchSession(ctx context.Context, initData *query.InitData) int {
 		failures = executeQueries(ctx, initData)
 	}
 	// set global exit code
-	return failures
+	return failures, nil
 }
 
 func executeQueries(ctx context.Context, initData *query.InitData) int {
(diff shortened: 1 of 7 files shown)


pkg/query/queryexecute/execute.go
```diff
@@ -18,31 +18,32 @@ import (
...
 )
 
-func RunInteractiveSession(ctx context.Context, initData *query.InitData) {
+func RunInteractiveSession(ctx context.Context, initData *query.InitData) error {
 	utils.LogTime("execute.RunInteractiveSession start")
 	defer utils.LogTime("execute.RunInteractiveSession end")
 
 	// the db executor sends result data over resultsStreamer
-	resultsStreamer, err := interactive.RunInteractivePrompt(ctx, initData)
+	result, err := interactive.RunInteractivePrompt(ctx, initData)
```
kaidaguerre, line 26: no keen on this - result has an error, but we also return an error, and we handle the 2 errors differently - this is a bit confusing.

also, if `err` is non nil we just fail - what is the exit code in this case?

kaidaguerre (changes requested):

pkg/query/queryexecute/execute.go
pskrbasu: fixed

pkg/query/queryexecute/execute.go
```diff
@@ -54,7 +55,7 @@ func RunBatchSession(ctx context.Context, initData *query.InitData) int {
 		failures = executeQueries(ctx, initData)
 	}
 	// set global exit code
```
kaidaguerre, line 57: is this comment outdated?

pkg/query/queryexecute/execute.go
```diff
@@ -18,31 +18,32 @@ import (
...
-func RunInteractiveSession(ctx context.Context, initData *query.InitData) {
+func RunInteractiveSession(ctx context.Context, initData *query.InitData) error {
 	utils.LogTime("execute.RunInteractiveSession start")
 	defer utils.LogTime("execute.RunInteractiveSession end")
 
 	// the db executor sends result data over resultsStreamer
-	resultsStreamer, err := interactive.RunInteractivePrompt(ctx, initData)
-	error_helpers.FailOnError(err)
+	result := interactive.RunInteractivePrompt(ctx, initData)
+	error_helpers.ShowError(ctx, result.PromptErr)
```
kaidaguerre, line 27: weren't we going to remove this?

kaidaguerre (changes requested):

pkg/query/queryexecute/execute.go
```diff
@@ -18,31 +18,32 @@ import (
...
+func RunBatchSession(ctx context.Context, initData *query.InitData) (int, error) {
 	// start cancel handler to intercept interrupts and cancel the context
 	// NOTE: use the initData Cancel function to ensure any initialisation is cancelled if needed
 	contexthelpers.StartCancelHandler(initData.Cancel)
 
 	// wait for init
 	<-initData.Loaded
 	if err := initData.Result.Error; err != nil {
-		error_helpers.FailOnError(err)
+		return -1, err
```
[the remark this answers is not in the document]
pskrbasu: since the func also returns an int, so I returned -1 in case of failures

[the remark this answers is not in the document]
kaidaguerre: I would just return 0, the zero value of int. As there is an error, we would expect the first return to be ignored

kaidaguerre (changes requested):

PR #3003, opened 2023-01-16
Fixes issue where steampipe notifies of available update even if plugin is updated

binaek:


pkg/task/display.go
```diff
@@ -5,71 +5,73 @@ import (
...
-func (r *Runner) hasNotifications() bool {
+func (r *Runner) hasAvailableVersion() bool {
 	utils.LogTime("Runner.hasNotifications start")
 	defer utils.LogTime("Runner.hasNotifications end")
-	return files.FileExists(filepaths.NotificationsFilePath())
+	return files.FileExists(filepaths.AvailableVersionsFilePath())
 }
 
-func (r *Runner) getNotifications() (*Notifications, error) {
+func (r *Runner) getAvailableVersions() (*AvailableVersionCache, error) {
```
kaidaguerre, line 62: is this function loading from cache only - which case maybe call it `loadCachedVersions`

pkg/task/pluginversioncheck.go
```diff
@@ -1,92 +1,82 @@
...
-	"time"
-
-	"github.com/turbot/steampipe/pkg/constants"
-	"github.com/turbot/steampipe/pkg/plugin"
-)
-
 // check if there is a new version
-func checkPluginVersions(ctx context.Context, installationID string) []string {
-	var notificationLines []string
+// func checkPluginVersions(ctx context.Context, installationID string) []string {
```
kaidaguerre, line 4: delete if unused

kaidaguerre (changes requested):

kaidaguerre (changes requested):
still some commented out code to remove

PR #3037, opened 2023-01-25
Add logs in test workflow to debug intermittent failures/stalls

pskrbasu:


pskrbasu:
https://github.com/turbot/steampipe/pull/3070

PR #3039, opened 2023-01-25
Add a helpful error message when snapshot sharing fails because of a corrupted/invalid token. Closes #2944

### LINKED ISSUE #2944: Running `steampipe ... --share` with an invalid/corrupted token gives an unhelpful error message
```
steampipe ✗ steampipe query "select 1 as val, 2 as col" --share
Error: 400 Bad Request
```

pskrbasu:
Screenshot 2023-01-25 at 6 14 24 PM

### DIFF
--- a/pkg/cloud/workspace.go
+++ b/pkg/cloud/workspace.go
@@ -16,11 +16,8 @@ import (
 func GetUserWorkspaceHandle(ctx context.Context, token string) (string, error) {
 	client := newSteampipeCloudClient(token)
 	actor, _, err := client.Actors.Get(ctx).Execute()
-	if err.Error() == "400 Bad Request" {
-		return "", fmt.Errorf("Invalid token.\nPlease run %s or setup a token.", constants.Bold(fmt.Sprintf("steampipe login")))
-	}
 	if err != nil {
-		return "", sperr.Wrap(err)
+		return "", sperr.WrapWithMessage(err, fmt.Sprintf("Invalid token.\nPlease run %s or setup a token.", constants.Bold("steampipe login")))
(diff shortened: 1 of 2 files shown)


kaidaguerre (commented):
How does this compare to the display of `MissingCloudTokenError`:
```
 MissingCloudTokenError = errors.New("Not authenticated for Steampipe Cloud.\nPlease run 'steampipe login' or setup a token.")
```

pskrbasu:
> How does this compare to the display of `MissingCloudTokenError`:
> 
> ```
>  MissingCloudTokenError = errors.New("Not authenticated for Steampipe Cloud.\nPlease run 'steampipe login' or setup a token.")
> ```

We get the `MissingCloudTokenError` when the user is not logged in. This PR handles the error when the cloud token stored in cloud.steampipe.io.sptt file gets corrupted somehow and we receive an unhelpful `400 Bad Request` error .

kaidaguerre:
> > How does this compare to the display of `MissingCloudTokenError`:
> > ```
> >  MissingCloudTokenError = errors.New("Not authenticated for Steampipe Cloud.\nPlease run 'steampipe login' or setup a token.")
> > ```
> 
> We get the `MissingCloudTokenError` when the user is not logged in. This PR handles the error when the cloud token stored in cloud.steampipe.io.sptt file gets corrupted somehow and we receive an unhelpful `400 Bad Request` error .

right but my point is the output should be consistent, i.e. if the command is bolded in one, it should be in both

pskrbasu:
https://github.com/turbot/steampipe/pull/3069

PR #3046, opened 2023-01-30
Fix query batch mode returning zero exit code when rows return errors. Closes #3044

### LINKED ISSUE #3044: Steampipe returns 0 exit code when a plugin fails to do "something"
**Describe the bug**
Similar to https://github.com/turbot/steampipe/issues/842. I use shell scripts to wrap up steampipe queries for reporting, and run them via CI/CD pipeline. The issue I am facing is, if a steampipe plugin e.g. AWS reports an error, the exit code from steampipe query is 0. This gives a false sense that the query completed successfully. I would expect that if any part of the query (or it's plugins fail) it should return a non-zero exit code. A use case for example is, a misconfiguration in aws.spc and an assume-role operation fails.

steampipe version 0.17.4

**To reproduce**
Install AWS plugin
Add some dummy config to aws.spec
Try and do a query against AWS
Observe the error similar to below (I am running on my local machine here):
```
Error: rpc error: code = DeadlineExceeded desc = operation error STS: GetCallerIdentity, failed to sign request: failed to retrieve credentials: failed to refresh cached credentials, no EC2 IMDS role found, operation error ec2imds: GetMetadata, canceled, context deadline exceeded (SQLSTATE HV000)
```

Observe the exit code is 0:
```
[host ~]$ echo $?
0
```

**Expected behavior**
A non-zero exit code is returned

pskrbasu:


cmd/query.go
```diff
@@ -135,7 +135,7 @@ func runQueryCmd(cmd *cobra.Command, args []string) {
 	}
 	defer initData.Cleanup(ctx)
 
-	var failures int
+	var failures, rowErrors int
```
kaidaguerre, line 138: is there a need to distinguish between `failures` and `rowErrors` - can't we just use a single var?

kaidaguerre (changes requested):

pkg/display/display.go
```diff
@@ -53,7 +53,8 @@ func EnableTiming() DisplayOption {
 }
 
 // ShowOutput displays the output using the proper formatter as applicable
-func ShowOutput(ctx context.Context, result *queryresult.Result, opts ...DisplayOption) {
+func ShowOutput(ctx context.Context, result *queryresult.Result, opts ...DisplayOption) int {
+	rowErr := 0
```
kaidaguerre, line 57: `rowErrors`

pkg/query/queryexecute/execute.go
```diff
@@ -88,23 +91,24 @@ func executeQueries(ctx context.Context, initData *query.InitData) int {
...
 	if err != nil {
-		return err
+		return err, 0
 	}
 
+	rowErr := 0 // get the number of rows that returned an error
 	// print the data as it comes
 	for r := range resultsStreamer.Results {
-		display.ShowOutput(ctx, r, display.ShowTimingOnOutput(constants.OutputFormatTable))
+		rowErr = display.ShowOutput(ctx, r, display.ShowTimingOnOutput(constants.OutputFormatTable))
```
kaidaguerre, line 107: rowErrors

kaidaguerre (changes requested):

PR #3056, opened 2023-02-01
Fix issue where snapshot creation was failing for command line queries in batch mode. Closes #2943

### LINKED ISSUE #2943: Snapshot creation failing for command line queries in batch mode
Fails: 
```
steampipe ✗ steampipe query "select 1 as val, 2 as col" --share
Error: invalid name 'local..command_line_query' passed to ParseResourceName
```
same with `--snapshot`

but this works:
```
steampipe ✗ steampipe query "query.static_query" --share
Snapshot uploaded to https://cloud.steampipe.io/user/....
```

pskrbasu:


cmd/query.go
```diff
@@ -312,7 +312,7 @@ func ensureQueryResource(name string, resolvedQuery *modconfig.ResolvedQuery, w
 	shortName := "command_line_query"
 
 	// this is NOT a named query - create the query using RawSql
-	q := modconfig.NewQuery(&hcl.Block{}, w.Mod, shortName).(*modconfig.Query)
+	q := modconfig.NewQuery(&hcl.Block{Type: "query"}, w.Mod, shortName).(*modconfig.Query)
```
kaidaguerre, line 315:
```
	q := modconfig.NewQuery(&hcl.Block{Type: modconfig.BLockTypeQuery}, w.Mod, shortName).(*modconfig.Query)
```

kaidaguerre (changes requested):

PR #3092, opened 2023-02-07
Fixes issue where autocomplete does not quote table names that need to be quoted. Closes #3065

### LINKED ISSUE #3065: autocomplete does not quote table names that need to be quoted
**Describe the bug**
Some table names (especially for dynamic tables) need to be quoted because they contain characters that are not legal postgres identifiers.  These appear in the autocomplete list , but they are not quoted when you select them so the auto-completed query does not actually work.  

 
All versions is assume, but definitely 0.18

**To reproduce**
You can reproduce this by adding a table name that requires quoting, then trying to use autocomplete.  The CSV plugin is useful for testing this, as is will use the file name as the table name.  Create a file like `MyFile.csv`, then use the autocomplete to create a query - the table name is not quoted:
``` 
> select * from MyFile
Error: relation "myfile" does not exist (SQLSTATE 42P01)
```

**Expected behavior**
table names that require quoting should be quoted when added by autocomplete

pskrbasu:


pkg/utils/string_slice.go
```diff
@@ -10,3 +13,15 @@ func UnquoteStringArray(stringArray []string) []string {
 	}
 	return res
 }
+
```
kaidaguerre, line 16: wrong file? this is for string slices?

pkg/interactive/autocomplete.go
```diff
@@ -94,7 +95,7 @@ func sanitiseTableName(strToEscape string) string {
 	tokens := helpers.SplitByRune(strToEscape, '.')
 	escaped := []string{}
 	for _, token := range tokens {
-		if strings.ContainsAny(token, " -") {
+		if strings.ContainsAny(token, " -") || utils.ContainsUpper(token) {
```
kaidaguerre, line 98: out of interest, why are we checking foe `" ="`

kaidaguerre: add comment
`If string contains upper case chars, escape it as Postgres by default converts to lower case'

kaidaguerre (changes requested):

PR #3113, opened 2023-02-15
Fixes issue where snapshot mode in query leads to duplicate rows in console/file output

binaek:


tests/acceptance/test_files/snapshot.bats
```diff
@@ -0,0 +1,73 @@
+load "$LIB_BATS_ASSERT/load.bash"
+load "$LIB_BATS_SUPPORT/load.bash"
+
+# These set of tests are skipped locally
+# To run these tests locally set the SPIPETOOLS_TOKEN env var.
+# These tests will be skipped locally unless the above env var is set.
```
binaek, line 6: `unless the below env`

tests/acceptance/test_files/snapshot.bats
```diff
@@ -0,0 +1,73 @@
...
+
+function setup() {
+  if [[ -z "${SPIPETOOLS_TOKEN}" ]]; then
+    skip
+  else
```
binaek, line 11: Do we need the `else` ?

PR #3120, opened 2023-02-16
Adds a mechanism for status updates during initialisation

binaek:


pkg/db/db_local/create_client.go
```diff
@@ -155,6 +157,7 @@ func createMaintenanceClient(ctx context.Context, port int) (*pgx.Conn, error) {
 	// We want to wait for a LONG time for this to complete
 	// Use the context that was given - since that is tied to os.Signal
 	// and can be interrupted
+	statushooks.SetStatus(ctx, "Waiting for recovery")
```
kaidaguerre, line 160: is this still needed? when we also have `"Database is recovering. This may take some time."`

pkg/interactive/interactive_client.go
```diff
@@ -459,7 +459,10 @@ func (c *InteractiveClient) getQuery(ctx context.Context, line string) *modconfi
 			c.cancelActiveQueryIfAny()
 		}()
 
-		statushooks.SetStatus(ctx, "Initializing...")
+		statushooks.SetStatus(ctx, c.initData.Status)
```
kaidaguerre, line 462: comments please

pkg/initialisation/init_data.go
```diff
@@ -21,6 +21,9 @@ import (
 )
 
 type InitData struct {
+	Status          string
```
kaidaguerre, line 24: comments - when is this status used? when is the callback function set?

kaidaguerre (changes requested):

PR #3160, opened 2023-02-27
Rationalising 'status hooks'

binaek:

PR #3178, opened 2023-03-02
Fixes issue where status could not be shown for refresh connection during service start/restart

binaek:


binaek:
This is part of ongoing #3160

PR #3301, opened 2023-04-04
Detect older versions of MacOS and warn that Steampipe does not support them. Closes #3256

### LINKED ISSUE #3256: Detect older versions of MacOS and warn that Steampipe does not support them
Steampipe appears to require recent versions of MacOS. We've had [reports in Slack](https://steampipe.slack.com/archives/C01UECB59A7/p1678895226632539) that it does not work with Mojave and earlier:

```
$ steampipe query
Welcome to Steampipe v0.19.2
For more information, type .help
> select * fro
Error: could not load library "/.../.steampipe/db/14.2.0/postgres/lib/postgresql/steampipe_postgres_fdw.so": dlopen(/.../.steampipe/db/14.2.0/postgres/lib/postgresql/steampipe_postgres_fdw.so, 10): no suitable image found.  Did find:
	/.../.steampipe/db/14.2.0/postgres/lib/postgresql/steampipe_postgres_fdw.so: cannot load 'steampipe_postgres_fdw.so' (load command 0x80000034 is unknown)
	/.../.steampipe/db/14.2.0/postgres/lib/postgresql/steampipe_postgres_fdw.so: cannot load 'steampipe_postgres_fdw.so' (load command 0x80000034 is unknown) (SQLSTATE 58P01)
$ uname -a
Darwin ... 18.7.0 Darwin Kernel Version 18.7.0: Tue Jun 22 19:37:08 PDT 2021; root:xnu-4903.278.70~1/RELEASE_X86_64 x86_64 i386 Darwin
```

A [history of MacOS releases](https://en.wikipedia.org/wiki/MacOS_version_history#Releases) is available on Wikipedia.

We use Postgres 14, which requires macOS 10.14 or later per https://www.postgresql.org/download/macosx/

It would be great if Steampipe could detect the OS kernel version and warn the user that it is unsupported / too old, similar to what we do for WSL v1.

pskrbasu:
Now running Steampipe on MacOS Mojave and other older versions throws the error message:

`Error: Steampipe requires MacOS 10.15 (Catalina) and above, please upgrade and try again.`

### DIFF
--- a/main.go
+++ b/main.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/hashicorp/go-version"
 	_ "github.com/jackc/pgx/v5/stdlib"
 	"github.com/turbot/go-kit/helpers"
@@ -38,6 +39,9 @@ func main() {
 
 	// ensure steampipe is not run on WSL1
 	checkWsl1(ctx)
+
+	// check OSX kernel version
+	checkOSXVersion(ctx)
 
 	cmd.InitCmd()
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -110,3 +114,42 @@ func checkWsl1(ctx context.Context) {
 		}
 	}
 }
+
+func checkOSXVersion(ctx context.Context) {
+	// check if macOS
```
binaek, line 119: @pskrbasu Can we use a `runtime.GOOS` here?

binaek (changes requested):

Issue #3257, opened 2023-03-23
Including protocol in mod URLs for `steampipe mod install` should not throw an error


cbruno10:
**Is your feature request related to a problem? Please describe.**
When installing a mod `steampipe mod install`, if I copy and paste a URL from my browser for a repo, I get an error:
```
Error: 1 dependency failed to install - could not retrieve version data from Git URL 'https://github.com/turbot/steampipe-mod-aws-well-architected'
```

**Describe the solution you'd like**
If I don't pass a protocol, automatically use HTTPS and install

**Describe alternatives you've considered**
Trim the URL

**Additional context**

PR #3316, opened 2023-04-09
fix: trim protocol in url for mod commands

Mr-Destructive:
Remove `https` or `http` protocol in the URL as args in the mod commands
- Install
- Uninstall
- Update

Fix for #3257 

Fixes the error:
From
```bash
$ steampipe mod install https://github.com/turbot/steampipe-mod-hackernews-insights
Error: 1 dependency failed to install - could not retrieve version data from Git URL 'https://github.com/turbot/steampipe-mod-hackernews-insights'
```
To this;

```bash
$ steampipe mod install https://github.com/turbot/steampipe-mod-hackernews-insights

Installed 1 mod:

local
└── github.com/turbot/steampipe-mod-hackernews-insights@v0.3.0

```

It can also be done with simple string trimming like `opts.ModArgs[i] = opts.ModArgs[i][8:]` ,
but regex might be better for handling other protocols if needed.

Let me know if it's a valid fix or for any changes are required.

### DIFF
--- a/cmd/mod.go
+++ b/cmd/mod.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -90,6 +91,7 @@ func runModInstallCmd(cmd *cobra.Command, args []string) {
 
 	// if any mod names were passed as args, convert into formed mod names
 	opts := newInstallOpts(cmd, args...)
+	trimGitUrls(opts)
 	installData, err := modinstaller.InstallWorkspaceDependencies(ctx, opts)
 	error_helpers.FailOnError(err)
 
@@ -125,6 +127,7 @@ func runModUninstallCmd(cmd *cobra.Command, args []string) {
 	}()
 
 	opts := newInstallOpts(cmd, args...)
+	trimGitUrls(opts)
 	installData, err := modinstaller.UninstallWorkspaceDependencies(ctx, opts)
 	error_helpers.FailOnError(err)
 
@@ -160,7 +163,7 @@ func runModUpdateCmd(cmd *cobra.Command, args []string) {
 	}()
 
 	opts := newInstallOpts(cmd, args...)
-
+	trimGitUrls(opts)
 	installData, err := modinstaller.InstallWorkspaceDependencies(ctx, opts)
 	error_helpers.FailOnError(err)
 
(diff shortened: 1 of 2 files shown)


binaek:
Hi @Mr-Destructive ,

Thank you for the PR. As it so happens that this is actually an outstanding issue with `mod install`. I have linked the issue with your PR.

Can you please add a test which covers this functionality?

The test(s) should go in `[tests/acceptance/test_files/mod_install.bats] (https://github.com/Mr-Destructive/steampipe/blob/trim-git-url/tests/acceptance/test_files/mod_install.bats)`

tests/acceptance/test_files/mod_install.bats
```diff
@@ -74,6 +74,16 @@ local
   assert_output --partial 'does not satisfy mod.m4 which requires version 10.99.99'
 }
 
+@test "install a mod with protocal in url" {
+  steampipe mod install https://github.com/turbot/steampipe-mod-hackernews-insights@0.3.0
```
binaek, line 78: You need to prefix this statement with a `run` for `bats` to be able to pick up the output.

Otherwise, `assert_output` does not work.

binaek: @Mr-Destructive ^^^

cmd/mod.go
```diff
@@ -256,3 +259,14 @@ func newInstallOpts(cmd *cobra.Command, args ...string) *modinstaller.InstallOpt
 	}
 	return opts
 }
+
+// Modifies(trims) the URL if contains http ot https in arguments
+
+func trimGitUrls(opts *modinstaller.InstallOpts) {
+	re := regexp.MustCompile(`^(https?://)`)
```
binaek, line 266: @Mr-Destructive can we solve this with a `strings.HasPrefix` instead of a `regex` match (which is inherently greedy and expensive)

binaek (changes requested):

cmd/mod.go
```diff
@@ -263,10 +262,11 @@ func newInstallOpts(cmd *cobra.Command, args ...string) *modinstaller.InstallOpt
...
 
 func trimGitUrls(opts *modinstaller.InstallOpts) {
-	re := regexp.MustCompile(`^(https?://)`)
-	for i, url := range opts.ModArgs {
-		if re.MatchString(url) {
-			opts.ModArgs[i] = re.ReplaceAllString(url, "")
-		}
-	}
+    for i, url := range opts.ModArgs {
+        if strings.HasPrefix(url, "http://") {
```
binaek, line 266: `strings.TrimPrefix` already does a `HasPrefix` under-the-hood. This check is not required.

```
strings.TrimPrefix(url, "http://")
strings.TrimPrefix(url, "https://")
```

Mr-Destructive: Thank you, I didn't knew about that

PR #3328, opened 2023-04-12
Fixes issue where a mod would install even if plugin dependencies are not met

binaek:
depends on #3334

### DIFF
--- a/pkg/modinstaller/mod_installer.go
+++ b/pkg/modinstaller/mod_installer.go
@@ -184,10 +184,10 @@ func (i *ModInstaller) InstallWorkspaceDependencies(ctx context.Context) (err er
 	if !i.force {
 		// there's no point in checking the requirements if force is set
 		// since we will ignore it anyway
-		if err := workspaceMod.Require.ValidateSteampipeVersion(workspaceMod.Name()); err != nil && !i.force {
+		if err := workspaceMod.Require.ValidateSteampipeVersion(workspaceMod.Name()); err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


pkg/modinstaller/mod_installer.go
```diff
@@ -167,9 +181,15 @@ func (i *ModInstaller) InstallWorkspaceDependencies(ctx context.Context) (err er
 		}
 	}()
 
-	// first check our Steampipe version is sufficient
-	if err := workspaceMod.Require.ValidateSteampipeVersion(workspaceMod.Name()); err != nil {
-		return err
+	if !i.force {
```
kaidaguerre, line 184: should we return warnings if force is used?

pkg/modinstaller/mod_installer.go
```diff
@@ -241,10 +264,16 @@ func (i *ModInstaller) commitShadow() (err error) {
 
 func (i *ModInstaller) installMods(ctx context.Context, mods []*modconfig.ModVersionConstraint, parent *modconfig.Mod) (err error) {
 	defer func() {
-		if err == nil {
+		if !i.dryRun && err != nil && i.force {
```
kaidaguerre, line 267: messy logic.

how about 
```
        // ensure we commit any changes to the shadow directory
	// (unless this was a dry run)
	if i.dryRun {
		log.Printf("[TRACE] installMods with dry-run=true returning without committing")
	} else {
		// commit if there is no error (or if force is set)
		if err == nil || i.force {
			err = i.commitShadow(ctx)
		}
	}
```

kaidaguerre (changes requested):

Issue #2953, opened 2023-01-06
Add a --silent option for `steampipe plugin install`


jchrisfarris:
**Is your feature request related to a problem? Please describe.**
By default `steampipe plugin install` shows the download and install progress. When running this in a script or redirecting the output to stdout, you end up with data that looks like this:

```
[Container] 2023/01/06 21:14:40 Running command ./steampipe plugin install terraform
45 |  
46 | terraform            [=========>----------------------------------------------------------] Downloading
47 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
48 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
49 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
50 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
51 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
52 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
53 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
54 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
55 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
56 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
57 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
58 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
59 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
60 | ·[1Aterraform            [=========>----------------------------------------------------------] Downloading
<TRUNCATED>
136 | ·[1Aterraform            [=====================>----------------------------------------------] Installing Plugin
137 | ·[1Aterraform            [=====================>----------------------------------------------] Installing Plugin
<TRUNCATED>
151 | ·[1Aterraform            [=====================>----------------------------------------------] Installing Plugin
152 | ·[1Aterraform            [=====================>----------------------------------------------] Installing Plugin
153 | ·[1Aterraform            [=====================>----------------------------------------------] Installing Plugin
154 | ·[1Aterraform            [====================================================================] Done
155 |  
156 | Installed plugin: terraform@latest v0.3.0
157 | Documentation:    https://hub.steampipe.io/plugins/turbot/terraform
158
```

**Describe the solution you'd like**
A CLI parameter that would suppress the progress, similar to the `--progress=false` on the `steampipe check` command.

PR #3342, opened 2023-04-15
add silent flag for plugin install cmd

Mr-Destructive:
#2953 
Hello, I have tried to add a `--silent` flag for the plugin install command.

```sh
$ steampipe plugin install --silent docker
````
Screenshot from 2023-04-15 15-04-30

Though the shorthand doesn't work, it shows in the docs below but I can't pick that up. Any suggestions?

Screenshot from 2023-04-15 15-53-54

Please review and let me know of any changes.
Thank you

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -82,12 +82,16 @@ Examples:
   steampipe plugin install aws
 
   # Install a specific plugin version
-  steampipe plugin install turbot/azure@0.1.0`,
+  steampipe plugin install turbot/azure@0.1.0
+
+  # Install a plugin with silent/quiet progress indicator
+  steampipe plugin install --silent aws`,
 	}
 
 	cmdconfig.
 		OnCmd(cmd).
-		AddBoolFlag(constants.ArgHelp, false, "Help for plugin install", cmdconfig.FlagOptions.WithShortHand("h"))
+		AddBoolFlag(constants.ArgHelp, false, "Help for plugin install", cmdconfig.FlagOptions.WithShortHand("h")).
+		AddBoolFlag(constants.ArgSilent, false, "No installation progress indicator", cmdconfig.FlagOptions.WithShortHand("s"))
 	return cmd
 }
 
@@ -201,6 +205,10 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	// plugin names can be simple names ('aws') for "standard" plugins,
 	// or full refs to the OCI image (us-docker.pkg.dev/steampipe/plugin/turbot/aws:1.0.0)
 	plugins := append([]string{}, args...)
+	silent, err := cmd.Flags().GetBool("silent")
+	if err != nil {
+		error_helpers.ShowErrorWithMessage(ctx, err, fmt.Sprintf("Failed to parse silent flag '%v'", silent))
+	}
 	installReports := make(display.PluginInstallReports, 0, len(plugins))
 
 	if len(plugins) == 0 {
@@ -215,23 +223,22 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 
 	// a leading blank line - since we always output multiple lines
 	fmt.Println()
-
 	progressBars := uiprogress.New()
 	installWaitGroup := &sync.WaitGroup{}
 	dataChannel := make(chan *display.PluginInstallReport, len(plugins))
 
-	progressBars.Start()
-
+	if !silent {
+		progressBars.Start()
+	}
 	for _, pluginName := range plugins {
 		installWaitGroup.Add(1)
 		bar := createProgressBar(pluginName, progressBars)
-		go doPluginInstall(ctx, bar, pluginName, installWaitGroup, dataChannel)
+		go doPluginInstall(ctx, bar, pluginName, installWaitGroup, dataChannel, silent)
 	}
 	go func() {
 		installWaitGroup.Wait()
 		close(dataChannel)
 	}()
-
 	installCount := 0
 	for report := range dataChannel {
 		installReports = append(installReports, report)
@@ -239,8 +246,9 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 			installCount++
 		}
 	}
-
-	progressBars.Stop()
+	if !silent {
+		progressBars.Stop()
+	}
 
 	if installCount > 0 {
 		// TODO do we need to refresh connections here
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cmd/plugin.go [hunk 1]
```diff
@@ -82,12 +82,16 @@ Examples:
...
+
+  # Install a plugin with silent/quiet progress indicator
+  steampipe plugin install --silent aws`,
 	}
 
 	cmdconfig.
 		OnCmd(cmd).
-		AddBoolFlag(constants.ArgHelp, false, "Help for plugin install", cmdconfig.FlagOptions.WithShortHand("h"))
+		AddBoolFlag(constants.ArgHelp, false, "Help for plugin install", cmdconfig.FlagOptions.WithShortHand("h")).
+		AddBoolFlag(constants.ArgSilent, false, "No installation progress indicator", cmdconfig.FlagOptions.WithShortHand("s"))
```
binaek, line 94: Since we are only disabling the progress bars I think this should be `--progress` and not `--silent`. This is also inline with the `check` command which has a similar funcitonality

https://github.com/turbot/steampipe/blob/f47985f88fc111d267640afec062e67f09d62efa/cmd/check.go#L68

cmd/plugin.go
```diff
@@ -201,6 +205,10 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	// plugin names can be simple names ('aws') for "standard" plugins,
 	// or full refs to the OCI image (us-docker.pkg.dev/steampipe/plugin/turbot/aws:1.0.0)
 	plugins := append([]string{}, args...)
+	silent, err := cmd.Flags().GetBool("silent")
```
binaek, line 208: Is there any reason you are pulling the value from the cmdFlags as opposed to `viper` like we do in other places?

binaek: To put this in context, all flag values are extracted and put in a global key value map. This is done in the PreRun binding phase.

Example:
https://github.com/turbot/steampipe/blob/f47985f88fc111d267640afec062e67f09d62efa/cmd/check.go#L143

binaek: I am not sure yet, but I suspect this is why the short hand is not working

cmd/plugin.go
```diff
@@ -201,6 +205,10 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
...
+	if err != nil {
```
binaek, line 209: This `if block` should not be required.

cmd/plugin.go
```diff
@@ -215,32 +223,32 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
...
-	progressBars.Start()
-
+	if !silent {
+		progressBars.Start()
+	}
 	for _, pluginName := range plugins {
 		installWaitGroup.Add(1)
 		bar := createProgressBar(pluginName, progressBars)
-		go doPluginInstall(ctx, bar, pluginName, installWaitGroup, dataChannel)
+		go doPluginInstall(ctx, bar, pluginName, installWaitGroup, dataChannel, silent)
```
binaek, line 236: Do we need to send down the `silent` flag here? Can we just ignore the publications in the `dataChannel` if the flag is set? That will consolidate the `silent` logic in one place.

cmd/plugin.go
```diff
@@ -262,17 +270,19 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	fmt.Println()
 }
 
-func doPluginInstall(ctx context.Context, bar *uiprogress.Bar, pluginName string, wg *sync.WaitGroup, returnChannel chan *display.PluginInstallReport) {
+func doPluginInstall(ctx context.Context, bar *uiprogress.Bar, pluginName string, wg *sync.WaitGroup, returnChannel chan *display.PluginInstallReport, silent bool) {
```
binaek, line 273: We shouldn't need to send down the `silent` flag

binaek (changes requested):

pkg/constants/args.go
```diff
@@ -3,6 +3,7 @@ package constants
 // Argument name constants
 const (
 	ArgHelp                 = "help"
+	ArgSilent               = "silent"
```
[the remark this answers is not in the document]
Mr-Destructive: So, according to the current flag used in other places, does it needs to show progress without parsing the flag or do we need to use `--progress false` to hide the progress?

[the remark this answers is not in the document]
kaidaguerre: when currently defined, for the `check` command for example, the flag is defaulted to `true` so there is no need to pass it in normal use. 

To hide the progress you would pass `--progress=false`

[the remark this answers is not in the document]
binaek: Hi @Mr-Destructive , were you able to take a look at this and implement the changes?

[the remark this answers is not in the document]
Mr-Destructive: Sorry, I was on a 2-week break, I will update this quickly by today or tomorrow.

binaek (changes requested):
@Mr-Destructive thank you the swift updates. Requesting a couple of small tweaks. 

Also, I feel that we should also introduce a `--progress` flag to `steampipe plugin update` - similar behaviour and functionality. This is because `update` is very closely related to `install` and should replicate behaviour whenever possible.

cmd/plugin.go
```diff
@@ -215,13 +220,13 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 
 	// a leading blank line - since we always output multiple lines
 	fmt.Println()
-
 	progressBars := uiprogress.New()
 	installWaitGroup := &sync.WaitGroup{}
 	dataChannel := make(chan *display.PluginInstallReport, len(plugins))
```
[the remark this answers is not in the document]
Mr-Destructive: done

cmd/plugin.go
```diff
@@ -82,11 +82,15 @@ Examples:
...
-  steampipe plugin install turbot/azure@0.1.0`,
+  steampipe plugin install turbot/azure@0.1.0
+
+  # Install a plugin without progress indicator
+  steampipe plugin install --progress=false aws`,
 	}
 
 	cmdconfig.
 		OnCmd(cmd).
+		AddBoolFlag(constants.ArgProgress, true, "Display control execution progress").
```
binaek, line 93: Can you please update the description of the `--progress` flag. This seems to be the description used for `check`.

I think set it to `Display installation progress`.

cmd/plugin.go
```diff
@@ -117,6 +121,7 @@ Examples:
 	cmdconfig.
 		OnCmd(cmd).
 		AddBoolFlag(constants.ArgAll, false, "Update all plugins to its latest available version").
+		AddBoolFlag(constants.ArgProgress, true, "Display control execution progress").
```
binaek, line 124: Update description for plugin (see above)

cmd/plugin.go
```diff
@@ -82,11 +82,15 @@ Examples:
   steampipe plugin install aws
 
   # Install a specific plugin version
-  steampipe plugin install turbot/azure@0.1.0`,
+  steampipe plugin install turbot/azure@0.1.0
+
+  # Install a plugin without progress indicator
+  steampipe plugin install --progress=false aws`,
```
binaek, line 88: Please add a similar example for `update`?

cmd/plugin.go
```diff
@@ -82,11 +82,15 @@ Examples:
   steampipe plugin install aws
 
   # Install a specific plugin version
-  steampipe plugin install turbot/azure@0.1.0`,
+  steampipe plugin install turbot/azure@0.1.0
+
+  # Install a plugin without progress indicator
```
binaek, line 87: I think change this to `Hide progress bars during installation`

binaek (changes requested):

Issue #3369, opened 2023-04-26
Allow dashboard service startup timeout to be specified as an environment variabled 


taliesins:
**Describe the bug**
When running the dashboard in a container with a couple of mods (all the aws and kubernetes ones) it will very often time out with message `Error: dashboard server startup timed out`.

**Steampipe version (`steampipe -v`)**
Example: v0.19.4

**To reproduce**
Steps to reproduce the behavior (please include relevant code and/or commands).

Steps to install in container:
```
# Use a constant workspace directory that can be mounted to
WORKDIR /workspace

# create a workdir for all mods
RUN mkdir -p /workspace/mods

# disable auto-update
ENV STEAMPIPE_UPDATE_CHECK=false

# disable telemetry
ENV STEAMPIPE_TELEMETRY=none

# Run steampipe service once and stop it
RUN cd /workspace/mods && \
    steampipe service start --dashboard && \
    steampipe service stop --force && \ 
    rm -f /home/steampipe/.steampipe/internal/.passwd

# Install all plugin and mods
RUN cd /workspace/mods && \
    steampipe plugin install steampipe terraform aws kubernetes && \
    steampipe mod install github.com/turbot/steampipe-mod-aws-compliance && \
    steampipe mod install github.com/turbot/steampipe-mod-aws-insights && \
    steampipe mod install github.com/turbot/steampipe-mod-aws-perimeter && \
    steampipe mod install github.com/turbot/steampipe-mod-aws-tags && \
    steampipe mod install github.com/turbot/steampipe-mod-aws-thrifty && \
    steampipe mod install github.com/turbot/steampipe-mod-kubernetes-insights && \
    steampipe mod install github.com/turbot/steampipe-mod-kubernetes-compliance
```
Entrypoint:
```
steampipe service start --dashboard-listen=network --database-port=5432 --dashboard-port=9194 --dashboard --foreground
```

**Expected behavior**
Should be able to specify the dashboard timeout as an environment variable (consider doing this for the other hard coded values for timeouts as well) for people that want to show a lot on the dashboards.

**Additional context**
I believe that this is the same issue as #2434. I think that the timeout vaules should rather be read from environment variables and default to the values in the code if they are not specified. Forcing a hard coded timeout I think is the issue.

binaek:
[...]

PR #3373, opened 2023-04-27
Add a configurable timeout for the dashboard

taliesins:
When you have a lot of dashboards it may take more then 30 seconds to start up especially in a resource contained environment i.e. running it in a container in Kubernetes


taliesins:
This closes #3369

binaek:
Hi @taliesins 

Thank you for raising this PR. Apologies for the late response.

I think if we are going to make `dashboard-timeout` configurable, it should also be read in as a configuration value.

We should have a `start-timeout` in `options.dashboard`, similar to what we have in `options.database` (refer: https://github.com/turbot/steampipe/issues/3428)

The settings are deserialised to the `GlobalDashboard` data structure as defined in https://github.com/turbot/steampipe/blob/main/pkg/steampipeconfig/options/dashboard.go#L16

Please let me know if you need any more details.

PR #3465, opened 2023-05-23
Updated sample config file behaviour as per the proposed changes. Closes #3431

### LINKED ISSUE #3431: Proposed change to "sample" config file behavior

When there are changes to the config file options, they will not be reflected in the `default.spc` for existing installations

Current:
- Steampipe will copy a `default.spc` to the `.steampipe/config` dir if it does not exists
- The `default.spc` contains sample values, but it is completely commented out.  It is meant to "advertise" the options/capabilities.
- The problem with this approach is that the file is not updated once it has been copied.  When there are changes to the config file options, they will not be reflected in the `default.spc` for existing installations.
- We copy a sample `default.spc`, but not a `workspaces.spc`.  

Proposed:
- Instead of copying the file as `default.spc`, copy if as `default.spc.sample`.  This way the file will only ever be an example - Steampipe would not use it unless it was renamed.
- Overwrite `default.spc.sample` whenever it changes so that existing installations also see the new format/values as the options change
- Do the same for `workspaces.spc.sample`
- If possible, try to detect and clean up old `default.spc` files if they have not been modified (hash compare maybe?).

pskrbasu:


pkg/steampipeconfig/load_config.go
```diff
@@ -49,6 +50,24 @@ func LoadConnectionConfig() (*SteampipeConfig, *modconfig.ErrorAndWarnings) {
 
 func ensureDefaultConfigFile(configFolder string) error {
 	defaultConfigFile := filepath.Join(configFolder, defaultConfigFileName)
+	defaultConfigSampleFile := filepath.Join(configFolder, defaultConfigSampleFileName)
+	// if default.spc exists, and if the content of the default.spc and the default.spc.sample
+	// file are the same, remove the default.spc file.
+	if _, err := os.Stat(defaultConfigFile); err == nil {
+		if equal, _ := utils.AreFilesEqual(defaultConfigFile, defaultConfigSampleFile); equal {
```
binaek, line 57: don't ignore the error. INFO log it at least

pkg/steampipeconfig/workspace_profile_loader.go
```diff
@@ -19,7 +24,26 @@ type WorkspaceProfileLoader struct {
...
+	err := os.WriteFile(defaultWorkspaceSampleFile, []byte(constants.DefaultWorkspaceContent), 0755)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
 func NewWorkspaceProfileLoader(workspaceProfilePath string) (*WorkspaceProfileLoader, error) {
+	// write the workspaces.spc.sample file
+	err := os.MkdirAll(workspaceProfilePath, 0755)
```
binaek, line 39: move the `MkDirAll` to `ensureDefaultWorkspaceFile`

binaek (changes requested):

PR #3472, opened 2023-05-23
Adds 'steampipe_server_settings' table populated with server settings data during service startup. Closes #3462

### LINKED ISSUE #3462: `steampipe_server_settings` table holding the configuration values of the running server
* cache setting values
* steampipe version
* FDW version
* start time in UTC

binaek:


pkg/db/db_local/server_settings.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/jackc/pgx/v5"
+	"github.com/spf13/viper"
+	"github.com/turbot/steampipe/pkg/constants"
+	"github.com/turbot/steampipe/pkg/db/db_common"
+	"github.com/turbot/steampipe/pkg/utils"
+	"github.com/turbot/steampipe/pkg/version"
+)
+
+type ServerSettingKey string
```
kaidaguerre, line 17: does having this as a type really add anything? it's a column key, which intuitively is a string anyway?

pkg/db/db_local/server_settings.go
```diff
@@ -0,0 +1,121 @@
...
+		// this also allows us to modify the table structure without having to go through complex
+		// migrations
+		getServerSettingsTableDropSQL(ctx),
+		// create a new one
+		getServerSettingsTableCreateSQL(ctx),
+		// grants
+		getServerSettingsTableGrantSQL(ctx),
+	}
+
+	queries = append(queries, getServerSettingsRowSql(ctx, settings)...)
```
kaidaguerre, line 57: why append this separately?

pkg/db/db_local/server_settings.go
```diff
@@ -0,0 +1,121 @@
...
+
+	_, err := ExecuteSqlWithArgsInTransaction(ctx, conn, queries...)
+	return err
+}
+
+func getServerSettingsRowSql(_ context.Context, settings map[ServerSettingKey]any) []db_common.QueryWithArgs {
+	queries := []db_common.QueryWithArgs{}
+	for name, value := range settings {
+		dataType := "text"
+		kind := reflect.TypeOf(value).Kind()
```
kaidaguerre, line 67: yeah this is pretty nasty - instead can we json serialize and write that as jsonb? There must be a nice pattern for writing data to a jsonb column?

kaidaguerre: I think rather than this we should just have something which knows how to read/write the server settings, and avoid the need for a dataytype column

kaidaguerre (changes requested):

pkg/db/db_local/server_settings.go
binaek: `getServerSettingsRowSql` returns a slice. the `append` is so that we can `spread` it out.

pkg/db/db_local/server_settings.go
binaek: Just limiting the keys that can be put in

pkg/db/db_client/db_client.go
```diff
@@ -80,9 +84,15 @@ func NewDbClient(ctx context.Context, connectionString string, onConnectionCallb
 		return nil, err
 	}
 
+	// load up the server settings
+	if err := client.loadServerSettings(ctx); err != nil {
+		client.Close(ctx)
```
kaidaguerre, line 89: maybe we should have a defer which checks the error and closes the client if the client is non nil and the error is not nil

pkg/db/db_common/server_settings.go
```diff
@@ -0,0 +1,190 @@
...
+		constants.InternalSchema,
+		constants.ServerSettingsTable,
+	)
+
+	return QueryWithArgs{Query: query}
+}
+
+const (
+	ServerSettingSteampipeVersion ServerSettingKey = "steampipe_version"
+	ServerSettingFdwersion        ServerSettingKey = "fdw_version"
```
kaidaguerre, line 185: typo

pkg/db/db_common/server_settings.go
```diff
@@ -0,0 +1,190 @@
...
+	settings := new(ServerSettings)
+	for rows.Next() {
+		var name ServerSettingKey
+		var value any
+		if err := rows.Scan(&name, &value); err != nil {
+			return nil, sperr.WrapWithMessage(err, "error reading row from %s.%s", constants.InternalSchema, constants.ServerSettingsTable)
+		}
+
+		switch name {
+		case ServerSettingSteampipeVersion:
```
kaidaguerre, line 51: why not use custom tags on the server setting struct

```
type ServerSettings struct {
	StartTime        time.Time `column:"fdw_version,time"`
	SteampipeVersion string `column:"steampipe_version,text"`
	FdwVersion       string      `column:"fdw_version,text"`
	CacheMaxTtl      int          `column:"cache_max_ttl,int"`
	CacheMaxSizeMb   int.     etc
	CacheEnabled     bool

}

```
Then when reading rows, you can iterate through the properties of ServerSettings to find a matching tag, the switch on the tag type to set the property value:

```
switch tagType {
  case "text":
       setProperty(p, value.(string))
 case "int":
       setProperty(p, value.(int))
}
etc
```

}

kaidaguerre: we do similar stuff for introspection columns

pkg/db/db_local/start_services.go
```diff
@@ -156,13 +156,23 @@ func postServiceStart(ctx context.Context, res *StartResult) error {
...
 	}
-	// ensure connection stat etable contains entries for all connections in connection config
+	// ensure connection state table contains entries for all connections in connection config
 	// (this is to allow for the race condition between polling connection state and calling refresh connections,
 	// which does not update the connection_state with added connections until it has built the ConnectionUpdates
 	if err := initializeConnectionStateTable(ctx, conn); err != nil {
 		return err
 	}
 
+	// initialize the server settings table
```
kaidaguerre, line 166: // initialize the server settings table
	// this table contains that configuration for this service instance
	// (we don't support file watching on service config file so the settings 
	//  remain valid for the lifetime of this instance)

pkg/db/db_common/server_settings.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	settings.loaded = true
+	return settings, nil
+}
+
+// Loaded returns a bool indicating whether settings data has been loaded.
+//
+// Use this for backward compatibility with pre 0.21 servers
+func (s *ServerSettings) Loaded() bool {
+	return s.loaded
```
kaidaguerre, line 87: remove loaded property

pkg/db/db_common/server_settings.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+// SetupSql returns the set of SQL statements to fully replace any existing
+// settings table with a new one and populates the values
+func (s *ServerSettings) SetupSql(ctx context.Context) []QueryWithArgs {
```
kaidaguerre, line 92: bad name - `GetSerializationSQL` ?

pkg/db/db_common/server_settings.go
```diff
@@ -0,0 +1,190 @@
...
+		// grants
+		getServerSettingsTableGrantSQL(ctx),
+	}
+
+	queries = append(queries, getServerSettingsRowSql(ctx, settings)...)
+
+	return queries
+}
+
+func getServerSettingsRowSql(_ context.Context, settings map[ServerSettingKey]any) []QueryWithArgs {
```
kaidaguerre, line 122: inconsistent Sql casing

kaidaguerre (changes requested):

pkg/db/db_local/server_settings.go
```diff
@@ -0,0 +1,28 @@
...
+
+// setupServerSettingsTable creates a new read-only table with information in the current
+// settings the service has been started with.
+//
+// The table also includes the CLI and FDW versions for reference
+func setupServerSettingsTable(ctx context.Context, conn *pgx.Conn) error {
+	settings := serversettings.ServerSettings{
+		SteampipeVersion: version.VersionString,
+		FdwVersion:       constants.FdwVersion,
+		StartTime:        time.Now().UTC(),
```
kaidaguerre, line 22: UTC or local time?

pkg/serversettings/converters.go
```diff
@@ -0,0 +1,29 @@
+package serversettings
+
+import (
+	"reflect"
+	"time"
+)
+
+// tryConvertFromTime tries to serialize the given value(v) into a RFC3339 formatted
+// string if the given field is a time.Time
+func tryConvertFromTime(field reflect.Value, v any) (_ string, converted bool) {
```
kaidaguerre, line 10: why named return val?

pkg/serversettings/load.go
```diff
@@ -0,0 +1,73 @@
...
+		value := reflect.ValueOf(settingValue)
+
+		for i := 0; i < reflectedSettings.NumField(); i++ {
+			tag := reflectedSettings.Type().Field(i).Tag.Get("setting_key")
+
+			if tag != settingName {
+				continue
+			}
+
+			log.Println("[INFO] serversetting: loading value of", settingName, "into field", tag)
```
kaidaguerre, line 46: don't use INFO logging in CLI

pkg/serversettings/load.go
```diff
@@ -0,0 +1,73 @@
...
+			field := reflectedSettings.Field(i)
+			value = tryConvert(field, value, tag)
+
+			reflectedSettings.Field(i).Set(value)
+		}
+	}
+	return settings, nil
+}
+
+func tryConvert(field reflect.Value, value reflect.Value, tag string) reflect.Value {
```
kaidaguerre, line 56: function comment

pkg/serversettings/setup.go
```diff
@@ -0,0 +1,162 @@
...
+	err = setupGrantsOnServerSettingsTable(ctx, tx)
+	if err != nil {
+		return err
+	}
+	err = populateServerSettingsTable(ctx, tx, s)
+	if err != nil {
+		return err
+	}
+
+	return tx.Commit(ctx)
```
kaidaguerre, line 52: I like committing in the defer

kaidaguerre (changes requested):

pkg/db/db_client/db_client.go
```diff
@@ -24,6 +25,10 @@ type DbClient struct {
 	connectionString string
 	pool             *pgxpool.Pool
 
+	// the settings of the server that this client is
+	// connected to
```
kaidaguerre, line 29: 2 lines really needed??

pkg/db/db_client/db_client.go
```diff
@@ -76,25 +81,53 @@ func NewDbClient(ctx context.Context, connectionString string, onConnectionCallb
...
 
+func (c *DbClient) loadServerSettings(ctx context.Context) error {
+	conn, _, err := c.GetDatabaseConnectionWithRetries(ctx)
+	if err != nil {
+		return err
+	}
+	defer conn.Release()
+	serverSettings, err := serversettings.Load(ctx, conn.Conn())
+	if err != nil {
+		if _, _, notFound := IsRelationNotFoundError(err); notFound {
```
kaidaguerre, line 121: comment - when would we expect this to happen?

pkg/db/db_local/server_settings.go
kaidaguerre: ?????

pkg/db/db_local/server_settings.go
```diff
@@ -0,0 +1,38 @@
...
+	settings := db_common.ServerSettings{
+		SteampipeVersion: version.VersionString,
+		FdwVersion:       constants.FdwVersion,
+		StartTime:        time.Now().UTC(),
+		CacheEnabled:     viper.GetBool(constants.ArgServiceCacheEnabled),
+		CacheMaxTtl:      viper.GetInt(constants.ArgCacheMaxTtl),
+		CacheMaxSizeMb:   viper.GetInt(constants.ArgMaxCacheSizeMb),
+	}
+
+	queries := []db_common.QueryWithArgs{}
```
kaidaguerre, line 29: can't this be: 
```
queries := []db_common.QueryWithArgs{
 serversettings.DropServerSettingsTable(ctx),
 serversettings.CreateServerSettingsTable(ctx),
 serversettings.GrantsOnServerSettingsTable(ctx),
 serversettings.GetPopulateServerSettingsSql(ctx, settings),
}

kaidaguerre (changes requested):

PR #3589, opened 2023-06-19
Fixes issue where aggregator connections are always getting updated. Closes #3582

### LINKED ISSUE #3582: Aggregators are always being considered requiring update when refresh connections is run
caused by the fact that `connections` property is not stored in connection state table, so we always detect a diff in this field

To fix add this column. Think about migration

binaek:


kaidaguerre (changes requested):
could we add an acceptance test for the migration?

pkg/db/db_local/internal.go
```diff
@@ -208,11 +208,15 @@ func validateFunction(f db_common.SQLFunction) error {
 }
 
 func initializeConnectionStateTable(ctx context.Context, conn *pgx.Conn) error {
-	// first create the table if necessary
-	createQueries := []db_common.QueryWithArgs{
-		connection_state.GetConnectionStateTableCreateSql(),
-		{Query: fmt.Sprintf(`GRANT SELECT ON TABLE %s.%s to %s;`, constants.InternalSchema, constants.ConnectionStateTable, constants.DatabaseUsersRole)},
-	}
+
+	createQueries := db_common.QueriesWithArgs{}.
```
kaidaguerre, line 212: no need for a new type - this is fine: 
```
	// create the table if necessary
	createQueries := []db_common.QueryWithArgs{
		connection_state.GetConnectionStateTableCreateSql(),
	}
	// add columns which were added later than 0.20.0
	createQueries = append(createQueries, connection_state.GetConnectionStateTableColumnAlterSql()...)

	// setup grants on the table
	createQueries = append(createQueries, connection_state.GetConnectionStateGrantSql())
```

pkg/connection_state/connection_state_table_sql.go
```diff
@@ -2,30 +2,66 @@ package connection_state
...
 import (
 	"fmt"
+
 	"github.com/turbot/steampipe/pkg/constants"
 	"github.com/turbot/steampipe/pkg/db/db_common"
 	"github.com/turbot/steampipe/pkg/steampipeconfig"
 	"github.com/turbot/steampipe/pkg/steampipeconfig/modconfig"
 )
 
+// extraColumnsWithType is a map of column names to the SQL type
```
kaidaguerre, line 12: I think define this inside `GetConnectionStateTableColumnAlterSql` function

kaidaguerre (changes requested):

PR #3590, opened 2023-06-19
Fixes issue where 'check' exports even if the run has been cancelled. Closes #3578

### LINKED ISSUE #3578: steampipe should not create export files for Cancelled control runs

binaek:


cmd/check.go
```diff
@@ -167,20 +167,22 @@ func runCheckCmd(cmd *cobra.Command, args []string) {
...
-		exportMsg, err = initData.ExportManager.DoExport(ctx, exportName, executionTree, exportArgs)
-		error_helpers.FailOnError(err)
-
-		// if the share args are set, create a snapshot and share it
-		if generateSnapshot {
-			err = controldisplay.PublishSnapshot(ctx, executionTree, shouldShare)
-			if err != nil {
-				exitCode = constants.ExitCodeSnapshotUploadFailed
-				error_helpers.FailOnError(err)
+		if !error_helpers.IsContextCanceled(ctx) {
```
kaidaguerre, line 170: how about putting this block into a function and having the contrext check at the top of the function - would clean up the top level code

binaek:
The command handler of the `check` command has become too big  and unmaintainable.

We can do the work in this PR as part of the refactor of the function

PR #3592, opened 2023-06-19
Fixes issue where exporting check all yields a badly formatted filename. Closes #3591

### LINKED ISSUE #3591: exporting `check all` yields a badly formatted filename
Screenshot 2023-06-19 at 7 27 12 PM

binaek:
Screenshot 2023-06-19 at 7 29 02 PM

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -202,13 +202,12 @@ func runCheckCmd(cmd *cobra.Command, args []string) {
 // getExportName resolves the base name of the target file
 func getExportName(targetName string, modShortName string) string {
 	parsedName, _ := modconfig.ParseResourceName(targetName)
-	exportName := ""
 	if targetName == "all" {
-		exportName = fmt.Sprintf("%s.%s", modShortName, parsedName.Name)
-	} else {
-		exportName = parsedName.ToFullNameWithMod(modShortName)
+		// there will be no block type = manually construct name
+		return fmt.Sprintf("%s.%s", modShortName, parsedName.Name)
 	}
-	return exportName
+	// default to just converting to valid resource name
+	return parsedName.ToFullNameWithMod(modShortName)
 }
 
 // get the exit code for successful check run
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -169,7 +169,12 @@ func runCheckCmd(cmd *cobra.Command, args []string) {
 
 		// add the mod name to the target name to get the export file root
 		parsedName, _ := modconfig.ParseResourceName(targetName)
-		exportName := parsedName.ToFullNameWithMod(w.Mod.ShortName)
+		exportName := ""
+		if targetName == "all" {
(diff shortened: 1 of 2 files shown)


pkg/steampipeconfig/modconfig/parsed_name.go
```diff
@@ -55,7 +55,12 @@ func (p *ParsedResourceName) ToFullNameWithMod(mod string) string {
 }
 
 func BuildFullResourceName(mod, blockType, name string) string {
```
kaidaguerre, line 57: why is this being called with an empty block type?  `mod.name` is not a valid resource name. 

I think this function shgould validate it's params but the calling code should possibly not use this function

kaidaguerre (changes requested):

cmd/check.go
```diff
@@ -202,6 +199,18 @@ func runCheckCmd(cmd *cobra.Command, args []string) {
 	exitCode = getExitCode(totalAlarms, totalErrors)
 }
 
+// getExportName resolves the base name of the target file
+func getExportName(targetName string, modShortName string) string {
+	parsedName, _ := modconfig.ParseResourceName(targetName)
```
kaidaguerre, line 204:
```
parsedName, _ := modconfig.ParseResourceName(targetName)
	if targetName == "all" {
             // there will be no block type -= manually construct name
             return fmt.Sprintf("%s.%s", modShortName, parsedName.Name)
	}
       // default to just converting to valid resource name
        return parsedName.ToFullNameWithMod(modShortName)
}

File: pkg/steampipeconfig/modconfig/parsed_name.go [hunk 3]
```diff
@@ -54,6 +54,7 @@ func (p *ParsedResourceName) ToFullNameWithMod(mod string) string {
 	return BuildFullResourceName(mod, p.ItemType, p.Name)
 }
 
+// BuildFullResourceName generates a Fully Qualified name from the given components
```
kaidaguerre, line 57: no need for caps  -> `fully qualified`

give example

kaidaguerre: either include validation here or raise issue for validation and put a TODO here referencing issue

kaidaguerre (changes requested):

kaidaguerre (changes requested):

pkg/steampipeconfig/modconfig/parsed_name.go
```diff
@@ -11,6 +11,11 @@ type ParsedResourceName struct {
 	Name     string
 }
 
+// ParseResourceName splits up a resource name into it's components
```
kaidaguerre, line 14: lovely - but I actually meant an example for BuildFullResourceName :)

kaidaguerre (changes requested):

pkg/steampipeconfig/modconfig/parsed_name.go
binaek: pushed

Issue #3535, opened 2023-06-07
All logs onto stdout?


mafrosis:
**Is your feature request related to a problem? Please describe.**

I run steampipe in docker as a service, and connect to it from a postgres client using a DSN style connection string, eg. `postgresql://steampipe:egg@steampipe:9193`.

Steampipe logs don't go to stdout, and so monitoring this service is hard. I have to `docker exec` and `tail` various different log files.

**Describe the solution you'd like**

And ENV var something like `ALL_LOGS_TO_STDOUT=true`

**Describe alternatives you've considered**

A docker hack:
https://stackoverflow.com/a/59363862

**Additional context**

If there's no obvious wild reason why we shouldn't do this, then I am happy to take a stab at implementing the feature

binaek:
Steampipe has three different log targets:
* CLI logs to stderr
* Postgresql logs (along with FDW logs) go into `~/.steampipe/logs/database-YYYY-MM-DD.log`
* Plugin processes log go into `~/.steampipe/logs/plugin-YYYY-MM-DD.log` (through the `plugin-manager`)

All of these are separate processes and hence have different `stdout` handles.

Merging all the logs into one place is possible, but I feel that it's unnecessary and sometimes counter-productive. Depending on the log level, each component can emit lots of log lines and therefore, it becomes very difficult to be able to parse through them when we are debugging.

binaek:
> CLI logs to stderr

We have plans to change this and have the CLI target log files by default - but log to `stderr` optionally (set an `ENV_VAR` or something like that)

mafrosis:
Thanks @binaek. I've been working through the steampipe source, and it's certainly a non-trivial change to achieve this. From your response, it also sounds like you are heading in the opposite direction 😄 

Standard practice when containerising is to log entirely on stdout/stderr. No files. https://12factor.net/logs

In kubernetes, we can work around apps who do not follow the best practice by doing something like:
https://kubernetes.io/docs/concepts/cluster-administration/logging/#sidecar-container-with-logging-agent

Based on my source code read through and your comment, I think the sidecar option is probably best! 

Let me know your thoughts, and I will close this issue if we agree

PR #3602, opened 2023-06-21
Add --single-log-file param to the service command

mafrosis:
Ref #3535 

This patch adds the `--single-log-file` parameter to `steampipe service`, which will then write plugin logs into `plugin.log` instead of the usual date filename eg. `plugin-2023-06-14.log`. Log file rotation is unaffected.

Likely issues to address before merge

* I'm not sure if the parameter name is particularly good; happy to replace any better suggestion 
* Tests - please let me know if you would like tests added on any particular interface


binaek:
@mafrosis thanks for raising the PR. I am not sure whether this flag is necessary, since the same thing can be achieved with 3rd party tooling around steampipe. (`cat ~/.steampipe/logs/plugin-*.log > plugin.log` ???)

johnsmyth:
@mafrosis Im surprised that you need a single file in order to tail and forward the logs. fluentbit can watch a directory, for example.  What are you using in the sidecar for log forwarding?

PR #3613, opened 2023-06-22
Fixes issue where connection state fails to load when connecting to pre-0.20.7 service. Closes #3612

### LINKED ISSUE #3612: `steampipe_connection_state` fails to load when connection to pre-0.20.7 service
This happens because `0.20.7` adds a `connections` column to the `steampipe_connection_state` table.

We need to handle connecting to services which do not have the new column in the connection state table.

binaek:


pkg/steampipeconfig/load_connection_state.go
```diff
@@ -85,23 +99,29 @@ func LoadConnectionState(ctx context.Context, conn *pgx.Conn, opts ...LoadConnec
...
+
+	query := buildLoadConnectionStateQuery(config)
 
+	log.Println("[TRACE] running query", query)
 	rows, err := conn.Query(ctx, query)
 	if err != nil {
+		// columns were added after the 0.20.0 release (connections for now)
+		// we need to handle the case where we are connected to an old version of
+		// service which doesn't have some of these columns
+		if column, isColumNotFound := isColumnNotFoundError(err); isColumNotFound {
```
kaidaguerre, line 117: change to 
```
if column, isColumNotFound := isColumnNotFoundError(err); isColumNotFound {
			// if this was not an added column, return the error as is
			if _, isAddedColumn := ConnectionStateTableAddedColumns[column]; isAddedColumn {
   			    // try to load with the added column ignored
			    return loadConnectionState(ctx, conn, ignoreColumns(append(config.ignoredColumns, column)...))
                         }
		}
```

kaidaguerre (changes requested):

PR #3338, opened 2023-04-14
Refactor display configuration options code

pskrbasu:

PR #3660, opened 2023-07-07
Add flag for disabling writing of default plugin config during plugin installation. Closes #3531. Closes #2206

### LINKED ISSUE #2206: Do not install plugin configuration by default
In https://github.com/francois2metz/setup-steampipe, I install steampipe plugins and configure them. The default behavior of plugin install is to copy the config file to the steampipe config directory. As I configure the plugin in JSON, I must remove the default one to make it works, see https://github.com/francois2metz/setup-steampipe/blob/daad2307a8a71a0933e66d22414bf904f0a1dfad/installer.js#L175.

A way (ie: a flag) to disable plugin config copy.

**Describe alternatives you've considered**
N/A.

**Additional context**
If you think this would be useful, I can take a look to implement it.

### LINKED ISSUE #3531: Add flag to allow disable creating plugin `.spc` file after installing plugin
When installing a plugin, I'd like to not have its default `.spc` created in my Steampipe config directory

Provide a flag that can be used with `steampipe plugin install`

**Describe alternatives you've considered**
Deleting the default plugin `.spc` that gets created

**Additional context**

Mr-Destructive:
Addressing #3531 

- Have added a `--default-config` flag on `plugin install` command, it is set to `true` by default.
- To omit the setting of default config for the installed plugin the user needs to set it to false.

```
steampipe plugin install --default-config=false supabase
```

The approach is to join the path of the default config dir i.e. `~/.steampipe/config/` and the plugin name with the `.spc` extension. Is there a better way to obtain the file path of the plugin config?
Let me know if this approach is valid, or is it a bit hacky?

Thank you.

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -546,7 +546,7 @@ func installPlugin(ctx context.Context, pluginName string, isUpdate bool, bar *u
 		}
 	}()
 
-	image, err := plugin.Install(ctx, pluginName, progress, ociinstaller.WithSkipConfigEnabled(viper.GetBool(constants.ArgSkipConfig)))
+	image, err := plugin.Install(ctx, pluginName, progress, ociinstaller.WithSkipConfig(viper.GetBool(constants.ArgSkipConfig)))
 	if err != nil {
 		msg := ""
 		_, name, stream := ociinstaller.NewSteampipeImageRef(pluginName).GetOrgNameAndStream()
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -19,6 +21,7 @@ import (
 	"github.com/turbot/steampipe/pkg/db/db_local"
 	"github.com/turbot/steampipe/pkg/display"
 	"github.com/turbot/steampipe/pkg/error_helpers"
+	"github.com/turbot/steampipe/pkg/filepaths"
 	"github.com/turbot/steampipe/pkg/installationstate"
 	"github.com/turbot/steampipe/pkg/ociinstaller"
 	"github.com/turbot/steampipe/pkg/ociinstaller/versionfile"
@@ -109,12 +112,16 @@ Examples:
   steampipe plugin install turbot/azure@0.1.0
 
   # Hide progress bars during installation
-  steampipe plugin install --progress=false aws`,
+  steampipe plugin install --progress=false aws
+
+  # Delete default plugin config file
+  steampipe plugin install --default-config=false aws`,
 	}
 
 	cmdconfig.
 		OnCmd(cmd).
 		AddBoolFlag(constants.ArgProgress, true, "Display installation progress").
+		AddBoolFlag(constants.ArgDefaultConfig, true, "Use default config file for plugin").
 		AddBoolFlag(constants.ArgHelp, false, "Help for plugin install", cmdconfig.FlagOptions.WithShortHand("h"))
 	return cmd
 }
@@ -237,6 +244,7 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 	// or full refs to the OCI image (us-docker.pkg.dev/steampipe/plugin/turbot/aws:1.0.0)
 	plugins := append([]string{}, args...)
 	showProgress := viper.GetBool(constants.ArgProgress)
+	defaultConfig := viper.GetBool(constants.ArgDefaultConfig)
 	installReports := make(display.PluginInstallReports, 0, len(plugins))
 
 	if len(plugins) == 0 {
@@ -293,6 +301,14 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 		}
 
 		statushooks.Done(ctx)
+	}
+	// if --default-config is set to false
+	if !defaultConfig {
+		//remove the config file
+		for _, plugin := range plugins {
+			configFilePath := filepath.Join(filepaths.EnsureConfigDir(), fmt.Sprintf("%s.spc", plugin))
+			os.Remove(configFilePath)
+		}
 	}
 	display.PrintInstallReports(installReports, false)
 
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -5,8 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 	"time"
@@ -21,7 +19,6 @@ import (
 	"github.com/turbot/steampipe/pkg/db/db_local"
 	"github.com/turbot/steampipe/pkg/display"
 	"github.com/turbot/steampipe/pkg/error_helpers"
-	"github.com/turbot/steampipe/pkg/filepaths"
 	"github.com/turbot/steampipe/pkg/installationstate"
 	"github.com/turbot/steampipe/pkg/ociinstaller"
 	"github.com/turbot/steampipe/pkg/ociinstaller/versionfile"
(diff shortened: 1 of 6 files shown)


binaek:
Hi @Mr-Destructive 

Thank you for the contribution. It looks like we are still to finalise on how to approach the problem that @cbruno10 raised. 

We will pick this up at the earliest.

While we 'zero' down on it, can you add the tests to verify that the functionality works. This will ensure that changes in the area don't end up breaking the feature.

The test file should be `service_and_plugin.bats`.

Mr-Destructive:
Sure, will add those👍

cmd/plugin.go
```diff
@@ -294,6 +302,14 @@ func runPluginInstallCmd(cmd *cobra.Command, args []string) {
 
 		statushooks.Done(ctx)
 	}
+	// if --default-config is set to false
+	if !defaultConfig {
```
pskrbasu, line 306: This is not the right way to solve this issue.

Consider this scenario:
1. steampipe plugin install aws - (aws.spc file gets created)
2. configure aws.spc with creds and query
3. steampipe plugin uninstall aws - (plugin gets uninstalled, but aws.spc stays on)
4. steampipe plugin install aws ----default-config=false - (would delete the existing aws.spc config)

We should not remove the config but rather prevent its creation in the first place.

pkg/constants/args.go
```diff
@@ -18,6 +18,7 @@ const (
 	ArgDashboard               = "dashboard"
 	ArgDashboardListen         = "dashboard-listen"
 	ArgDashboardPort           = "dashboard-port"
+	ArgDefaultConfig           = "default-config"
```
pskrbasu, line 21: rename it to `skip-config`
```suggestion
	ArgSkipConfig           = "skip-config"
```

pskrbasu (changes requested):
@Mr-Destructive requested some changes. Please take a look.

Here's a small gist of how to achieve this:

- The func [installPlugin](https://github.com/turbot/steampipe/blob/854819f8f141267348ea1b9f6aaec28e1b959a50/cmd/plugin.go#L529) should accept WithSkipConfig() as an input, which would be passed down the stack till here in [InstallPlugin](https://github.com/turbot/steampipe/blob/854819f8f141267348ea1b9f6aaec28e1b959a50/pkg/ociinstaller/plugin.go#L53).
- We would then validate whether WithSkipConfig is true or false using an if block in [L53](https://github.com/turbot/steampipe/blob/854819f8f141267348ea1b9f6aaec28e1b959a50/pkg/ociinstaller/plugin.go#L53)
- Based on that the config file will either be written or not written.

Take a look at https://github.com/turbot/steampipe/pull/3338/files. This has a similar structure. We want to design it exactly like this.

Let us know if you need any help with this.

Mr-Destructive:
Are the changes fine? It will require changing the signature of the InstallPlugins in the `ociinstaller/plugin`.  

Other changes I need to confirm:
- Moving the `configFileOption` to a separate file?
- Can't the `--skip-config` be directly checked at `ociiinstaller/plugin` level function to call from `viper.GetBool(constants.ArgsSkipConfig)` (is this a bad practice or it won't work?)

Let me know if I am going wrong here. Thank you

cmd/plugin.go
```diff
@@ -47,6 +47,25 @@ type pluginJsonOutput struct {
 	Warnings  []string          `json:"warnings"`
 }
 
+type configFileOption struct {
```
pskrbasu, line 50:
```suggestion
type pluginInstallConfig struct {
```

pskrbasu, line 50:
```suggestion
type pluginInstallConfig struct {
```
Move these into ociinstaller package

cmd/plugin.go
```diff
@@ -47,6 +47,25 @@ type pluginJsonOutput struct {
...
+	skip bool
```
pskrbasu, line 51:
```suggestion
	skipConfigFile bool
```

pskrbasu (changes requested):
@Mr-Destructive please take a look at the changes requested and get back if you need help.

To answer your questions:
> - Moving the configFileOption to a separate file?

Yes, move it to a separate file in ociinstaller package. It should not stay in cmd. (Move L50-67 in cmd/plugin.go to ociinstaller package)

> - Can't the --skip-config be directly checked at ociiinstaller/plugin level function to call from viper.GetBool(constants.ArgsSkipConfig) (is this a bad practice or it won't work?)

The simplest answer is it can 😁. But we don't want to do it that way. Our code goes through 3levels of packages here - the top level being `cmd`, then it goes into `plugin` and finally into `ociinstaller`. The `--skip-config` is passed at the topmost level. So checking it in the 3rd level would mean there's no existence of it in the levels between(which is not clean code). Hence we want to pass it down as options in every single level to make the code better.

> Also, should I add the test cases for this scenario?

Yes that is a valid scenario and it would be great if you can add a test to replicate that. Thanks

cmd/plugin.go
```diff
@@ -47,6 +47,25 @@ type pluginJsonOutput struct {
...
+}
+
+func NewConfigFileOption() *configFileOption {
+	skip := cmdconfig.Viper().GetBool(constants.ArgSkipConfig)
+	return &configFileOption{
+		skip: skip,
+	}
+}
+
+type SkipConfigFile = func(config *configFileOption)
```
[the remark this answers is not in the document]
pskrbasu: Move these into ociinstaller

Mr-Destructive:
Thank you for the detailed help. Have made the changes and added the test.

Is the test valid?  Let me know of any further changes.
Thank you

pkg/ociinstaller/plugin.go
```diff
@@ -34,6 +34,11 @@ func InstallPlugin(ctx context.Context, imageRef string, sub chan struct{}) (*St
 
 	ref := NewSteampipeImageRef(imageRef)
 	imageDownloader := NewOciDownloader()
+	config := &pluginInstallConfig{}
+	for _, opt := range opts {
+		opt(config)
+	}
+	skipPluginConfig := config.skipConfigFile
```
pskrbasu, line 41: These lines should be the first lines of the func. (Not a biggie nor it is wrong, it's just how we have done it elsewhere 😄 )

Having another variable here seems excessive. You don't need another variable here `skipPluginConfig` (just use `config.skipConfigFile` wherever needed)

tests/acceptance/test_files/service_and_plugin.bats
```diff
@@ -691,6 +691,61 @@ load "$LIB_BATS_SUPPORT/load.bash"
...
+
+  run test -f $tmpdir/config/aws.spc
+  assert_failure
+
+  rm -rf $tmpdir
+}
+
+@test "verify that plugin installed with --skip-config as false, should have default config .spc file in config folder" {
+  tmpdir=$(mktemp -d)
+  run steampipe plugin install aws --skip-config=false --install-dir $tmpdir
```
pskrbasu, line 707: --skip-config by default is false. No need to add the flag here.

pskrbasu (changes requested):
@Mr-Destructive Thank you for making the changes. The tests look good.

A few minor changes are requested. Thanks!

pkg/ociinstaller/config.go
```diff
@@ -0,0 +1,13 @@
+package ociinstaller
+
+type pluginInstallConfig struct {
+	skipConfigFile bool
+}
+
+type PluginInstallOption = func(config *pluginInstallConfig)
+
+func WithSkipConfigEnabled(skipConfigFile bool) PluginInstallOption {
```
pskrbasu, line 9:
```suggestion
func WithSkipConfig(skipConfigFile bool) PluginInstallOption {
```

pskrbasu (changes requested):
@Mr-Destructive The changes look good.
Just one minor change is requested. Please take a look and rebase with the latest main(has conflicts). 

Thanks!

binaek:
@Mr-Destructive could you also rebase your branch

PR #3680, opened 2023-07-18
Adds feature for the Steampipe CLI to log to files like plugin manager and postgres. Closes #2916

### LINKED ISSUE #2916: CLI should log to a file instead of the console
Currently, when we set the log level to `TRACE`, sends out it's log to the console. While having the log is useful for debugging, it is cumbersome and unwieldy. Instead, steampipe should always send the log out to a dedicated log file like the `plugin-manager` and `service`

binaek:
Write logs from the CLI to `$INSTALL_DIR/logs/steampipe-yyyy-mm-dd.log`

### DIFF
--- a/pkg/constants/env.go
+++ b/pkg/constants/env.go
@@ -33,6 +33,9 @@ const (
 	EnvTelemetry                = "STEAMPIPE_TELEMETRY"
 	EnvIntrospection            = "STEAMPIPE_INTROSPECTION"
 	EnvWorkspaceProfileLocation = "STEAMPIPE_WORKSPACE_PROFILES_LOCATION"
+	EnvDiagnostics              = "STEAMPIPE_DIAGNOSTICS"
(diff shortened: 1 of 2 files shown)


pkg/constants/env.go
```diff
@@ -33,6 +33,8 @@ const (
 	EnvTelemetry                = "STEAMPIPE_TELEMETRY"
 	EnvIntrospection            = "STEAMPIPE_INTROSPECTION"
 	EnvWorkspaceProfileLocation = "STEAMPIPE_WORKSPACE_PROFILES_LOCATION"
+	EnvDiagnostics              = "STEAMPIPE_DIAGNOSTICS"
```
kaidaguerre, line 36: remove

kaidaguerre (commented):
is this tested with log levels set in various places? i.e. testing `logLevelNeedsReset`

pkg/constants/env.go
binaek: done

PR #3688, opened 2023-07-26
Fixes issue where using a bad workspace-database with a valid token gives invalid token as the error. Closes #3610

### LINKED ISSUE #3610: Using a bad workspace database with a valid token gives `invalid token` as the error
Screenshot 2023-06-22 at 4 10 54 PM

pskrbasu:


pkg/cloud/cloud_metadata.go
```diff
@@ -35,7 +35,9 @@ func GetCloudMetadata(ctx context.Context, workspaceDatabaseString, token string
 		cloudWorkspace, _, err = client.OrgWorkspaces.Get(ctx, identityHandle, workspaceHandle).Execute()
 	}
 
-	if err != nil {
+	if err != nil && err.Error() == "404 Not Found" {
```
kaidaguerre, line 38: can we put an function in for these - where do we have similar functions - error_helpers?

kaidaguerre (changes requested):
can we

PR #3690, opened 2023-07-26
Adds an exhaustive list of "safe" sprig functions to the check templates. Closes #3661.

### LINKED ISSUE #3661: Enable all "safe" functions from sprig so that they are available in check output templates
Right now, we have a very short list of functions which are made available to `check` output templates during rendering.

However, having a small set also means that the list needs to be extended whenever something new comes up. This creates a loop of debates on deciding which functions are worthy and which are not - essentially creating a bottleneck for the template author looking for specific features.

Enabling all "safe" functions at the outset allows faster iteration for template author.

This is an extension of https://github.com/turbot/steampipe/pull/3638

> We define "safe" as functions which operate only on input provided to it and DOES NOT access the environment. 
>
> **Example**: All FileSystem functions and Network functions are **UNSAFE**.
>
> Though `crypto` functions fall in the same category as "safe" functions, they are inherently CPU hungry. Moreover, `crypto` functions have no defined use case as far as `check` templates are concerned.

binaek:


johnsmyth:
After speaking with @e-gineer I think we should shelve this for now.  Its not clear that it provides sufficient value to justify the additional weight.  Additionally, we would need to document all the functions that are included and it is still not clear to me which fuctions would be included in this PR

PR #3758, opened 2023-08-21
Fixes issue where temporary tables are lost in the middle of a long running interactive session. Closes #3543

### LINKED ISSUE #3543: Database connection should not be recycled when running in interactive mode
Currently, steampipe clients create a connection pool which recycles connections periodically. This is essentially to ensure that we don't get memory leaks.

However, in the interactive prompt, this is a problem when temporary tables are used. Since `TEMPORARY` tables are scoped to a connection (or a transaction), recycling the connection in the interactive prompt also means that we lose any `TEMPORARY` tables in the prompt.

binaek:


pkg/db/db_client/db_client.go
```diff
@@ -30,6 +30,9 @@ type DbClient struct {
 	// connection used to run system/plumbing queries (connection state, server settings)
 	managementPool *pgxpool.Pool
 
+	// a connection hijacked from the pool before the pool is disabled
+	hijackedConnection *pgx.Conn
```
kaidaguerre, line 34: rename var  and improve comment - why do we have a hijacked connection?
maybe call it `userConnection` ?

pkg/db/db_client/db_client.go
```diff
@@ -122,6 +125,20 @@ func (c *DbClient) closePools() {
 	c.managementPool.Close()
 }
 
+func (c *DbClient) DisablePool(ctx context.Context) error {
```
kaidaguerre, line 128: comment - what is this function used for - why do we disable the pool

pkg/db/db_client/db_client_session.go
```diff
@@ -28,13 +28,21 @@ func (c *DbClient) AcquireSession(ctx context.Context) (sessionResult *db_common
...
 
-	// get a database connection and query its backend pid
-	// note - this will retry if the connection is bad
-	databaseConnection, err := c.userPool.Acquire(ctx)
-	if err != nil {
-		sessionResult.Error = err
-		return sessionResult
+	var databaseConnection db_common.Releasable
+
+	if c.hijackedConnection == nil {
```
kaidaguerre, line 33: comment - what are we doing here????

```
// if we are using a single user connection, we will use this for the session
// NOTE: we must convert the connection into a `Releasable` interface so we can call `Release` on it when we are done
if c.hijackedConnection != nil {
    databaseConnection = db_common.ReleasableFromConn(c.hijackedConnection)
} else {
  // we are using a connection pool - retrieve a connection from the pool
  var err error
		databaseConnection, err = c.userPool.Acquire(ctx)
		if err != nil {
			sessionResult.Error = err
			return sessionResult
		}
 }
```

kaidaguerre (changes requested):
comments need work

PR #3784, opened 2023-08-31
Fixes issue where connections would get dropped in the interactive prompt. Closes #3781

### LINKED ISSUE #3781: Interactive prompt should use a connection pool limited to 1 persistent connection

binaek:


kaidaguerre (changes requested):

pkg/interactive/interactive_client.go
```diff
@@ -811,5 +811,6 @@ func (c *InteractiveClient) handleConnectionUpdateNotification(ctx context.Conte
 	c.executionLock.Lock()
 	defer c.executionLock.Unlock()
 
-	c.client().RefreshSessions(ctx)
+	session := c.client().RefreshSessions(ctx)
+	session.Session.Close(true)
```
kaidaguerre, line 815: what is this doing? why is it needed>

PR #3842, opened 2023-09-13
Adds support for installing all referenced plugins when no arguments are given to 'plugin install'. Closes #3451

### LINKED ISSUE #3451: `steampipe plugin install` without arguments should discover required plugins and install
This is similar to behaviour in `steampipe mod install` which installs the entire dependency tree.

binaek:


pkg/ociinstaller/imageref.go
```diff
@@ -95,16 +95,36 @@ func (r *SteampipeImageRef) GetOrgNameAndStream() (string, string, string) {
...
 func (r *SteampipeImageRef) GetFriendlyName() string {
-	_, pluginName, pluginStream := r.GetOrgNameAndStream()
-	if pluginStream == DefaultImageTag {
-		return pluginName
-	} else {
-		return fmt.Sprintf("%s@%s", pluginName, pluginStream)
+	return getCondensedImageRef(r.DisplayImageRef())
+}
+
+func getCondensedImageRef(imagePath string) string {
```
kaidaguerre, line 109: shouldn't `imagePath` be `imageRef`?

Can we have a unit test suite please

kaidaguerre (changes requested):

PR #3924, opened 2023-09-28
Update help for 'steampipe plugin install'

binaek:
`steampipe plugin install` can now auto discover required plugins and install them. It doesn't need arguments anymore

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -103,6 +103,9 @@ is latest. The name is a required argument.
 
 Examples:
 
+	# Install all missing plugins that are specified in configuration files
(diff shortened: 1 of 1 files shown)


cmd/plugin.go
```diff
@@ -103,6 +103,9 @@ is latest. The name is a required argument.
 
 Examples:
 
+	# Install all missing plugins that are specified in configuration files
```
kaidaguerre, line 106: inconsistent spaces/tabbing

kaidaguerre (changes requested):

Issue #3927, opened 2023-09-28
Plugin list outputs nothing when no plugins are installed


pskrbasu:
When no plugins are installed, plugin list should output an empty table.

v0.19.5
```
binaek@Turingmbp steampipe % steampipe plugin list --install-dir ~/test_install
+------------------+---------+-------------+
| Installed Plugin | Version | Connections |
+------------------+---------+-------------+
+------------------+---------+-------------+
```

v0.20.12
```
binaek@Turingmbp steampipe % steampipe plugin list --install-dir ~/test_install
binaek@Turingmbp steampipe % 
```

PR #3936, opened 2023-09-29
fix plugin list empty table

Mr-Destructive:
Fixing #3927 

Currently the table is only printed when there is atleast one a plugin installed. 

Let me know if this is a valid fix or not, will make changes accordingly.

Thank you.


kaidaguerre:
thanks @Mr-Destructive 

I merged even with the failed tests as they were an existing known issue already fixed on main

PR #3959, opened 2023-10-14
Fixes issue where a query would return 'null' for an empty resultset when output is set to json in 'query'. Closes #3955

### LINKED ISSUE #3955: Empty results return null when using json output
Version: v0.21.1

When you perform a query with Steampipe and use the default table output, an empty result will still show an empty result like in this example:

```
> select * from steampipe_connection where plugin = 'random'
+------+-------+------+-------------+---------------+-------+--------+-----------------+-------------+-------------+--------------+---------------------+------------>
| name | state | type | connections | import_schema | error | plugin | plugin_instance | schema_mode | schema_hash | comments_set | connection_mod_time | plugin_mod_>
+------+-------+------+-------------+---------------+-------+--------+-----------------+-------------+-------------+--------------+---------------------+------------>
+------+-------+------+-------------+---------------+-------+--------+-----------------+-------------+-------------+--------------+---------------------+------------>
```

...but if you then change to json output and perform a query that returns no results, you get `null`, but I would expect to get an empty array `[]`?

```
> .output json
> select * from steampipe_connection where plugin = 'random'
null
> select * from aws_s3_bucket where name = 'dfjskdgfhfdjkgf'
null
```

Mr-Destructive:
Fixes #3955 

I have tried to fix the nil map initialization of the JSON output.

```
> .output json
> select * from steampipe_connection where plugin = 'random'
[]
> 
```
Let me know if it is a valid fix, or if any other workaround is required.

Thanks.


binaek (changes requested):
@Mr-Destructive thank you for the PR. The change looks good. Can you add an acceptance test for this gap?

Mr-Destructive:
Are the tests sufficient? Let me know if any other conditions are needed.
Thanks.

binaek:
@Mr-Destructive I updated the PR title to more closely match the conventions we use in our repositories.

PR #3960, opened 2023-10-14
Fix custom registries bugs

francois2metz:
- the documentation link was not correct
- the uninstall message was not correct
- checking requirements on mods was always failing

## When installing ghcr.io/francois2metz/scalingo:

Before:
> Installed plugin: scalingo@latest v0.18.0
> Documentation:    https://hub.steampipe.io/plugins/ghcr.io/scalingo

After: 
> Installed plugin: scalingo@latest v0.18.0
>Documentation:    https://ghcr.io/francois2metz/scalingo

**I didn't updated the name of the plugin, but I believe the message scalingo@latest  could be misleading. What do you think ?**

## Uninstall

Before:
> Uninstalled plugin:
> * ghcr.io/scalingo

After:
> Uninstalled plugin:
> * ghcr.io/francois2metz/scalingo

## Mod requirement

A simple mod with this requirement would always fail regardless of the installation of the plugin
```
  require {
    plugin "ghcr.io/francois2metz/scalingo" {
      version = "0.18.0"
    }
  }
```

### DIFF
--- a/cmd/plugin.go
+++ b/cmd/plugin.go
@@ -578,6 +578,9 @@ func installPlugin(ctx context.Context, pluginName string, isUpdate bool, bar *u
 		versionString = " v" + image.Config.Plugin.Version
 	}
 	docURL := fmt.Sprintf("https://hub.steampipe.io/plugins/%s/%s", org, name)
+	if !image.ImageRef.IsFromSteampipeHub() {
+		docURL = fmt.Sprintf("https://%s/%s", org, name)
+	}
 	return &display.PluginInstallReport{
 		Plugin:         fmt.Sprintf("%s@%s", name, stream),
 		Skipped:        false,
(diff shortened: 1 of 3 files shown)


kaidaguerre:
Thanks for this @francois2metz , looks good.

> I didn't updated the name of the plugin, but I believe the message scalingo@latest could be misleading. What do you think 

There is a convention that when installing a plugin without specifying a version, it is installed as the `latest` stream. So `scalingo@latest` is the plugin stream which has been installed.

Why do you feel this is misleading? What would you suggest as a less misleading output?

francois2metz:
> Why do you feel this is misleading? What would you suggest as a less misleading output?

As the name is really important in the configuration file, i would suggest to have something like:
- `scalingo@latest v0.18.0` for turbot plugins installed from the hub
- `francois2metz/scalingo@latest v0.18.0` for francois2metz org installed from the hub
- `ghcr.io/francois2metz/scalingo@latest v0.18.0` when installed from ghcr

kaidaguerre:
> As the name is really important in the configuration file, i would suggest to have something like:

@francois2metz that sounds sensible - could you raise an issue to that effect?

Issue #3772, opened 2023-08-25
Add ability to install a specific mod branch/commit with `steampipe mod install`


cbruno10:
**Is your feature request related to a problem? Please describe.**
If I have some changes I'd like to test in a mod repository that are on a branch, but not in a tag yet, I'd like to be able to install it with `steampipe mod install`

**Describe the solution you'd like**
Be able to pass a branch name or commit ID to `steampipe mod install`

**Describe alternatives you've considered**
Create a temporary tag to pull down

**Additional context**

PR #3972, opened 2023-10-22
add option to install mod from branch ref

Mr-Destructive:
For #3772 

I have made changes to the `mod install` command to support branch references as Git URLs for installing a mod from a particular branch.

```
steampipe mod install https://github.com/turbot/steampipe-mod-aws-insights/tree/issue-iam
```

Before reviews, can I get a general sense of clarity if I am picking this one right? 

Some concerns to be addressed:
- Separation of version and branch tagging logic
- Versioning of mod after installing from a branch

Thanks.

### DIFF
--- a/pkg/modinstaller/mod_installer.go
+++ b/pkg/modinstaller/mod_installer.go
@@ -342,10 +342,14 @@ func (i *ModInstaller) installModDependencesRecursively(ctx context.Context, req
 		// short circuit if the execution context has been cancelled
 		return ctx.Err()
 	}
-	// get available versions for this mod
-	includePrerelease := requiredModVersion.Constraint.IsPrerelease()
-	availableVersions, err := i.installData.getAvailableModVersions(requiredModVersion.Name, includePrerelease)
-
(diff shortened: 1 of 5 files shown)


kaidaguerre:
Hi @Mr-Destructive 

> Before reviews, can I get a general sense of clarity if I am picking this one right?
yes, this issue is not a straightforward one - there are decisions to be made regarding how we approach installing a mod from a branch, i.e. do we freeze the dependency to the latest commit at installation time - or do we 'track' the branch and always pull the latest commit when doing a mod update.

We need to have some discussion on our chosen approach for this one so let me get back to you once we have some clarity and an agreed approach.

kaidaguerre:
@Mr-Destructive we have some rafactoring of the mod installation on the horizon, which means it is probably not the right time to proceed with this PR, so I am rejecting it.

However, once the changes have been finalised, and we have some clarity on how we want this behaviour to work, I'd certainly be open for you to take a look at it. Watch this space.

Mr-Destructive:
Sure 👍

PR #4016, opened 2023-12-10
Fix inconsistent value for plugin in steampipe_plugin_column table - always use long name.

kaidaguerre:


kaidaguerre:
alreeady merged via v0.21.x

PR #4031, opened 2023-12-18
Simplifies available port check. Closes #4030

### LINKED ISSUE #4030: Simplify open port check
`pkg/utils.IsPortBindable` uses `DialTimeout` to check for port availability. This can be simplified with `net.Listen`

binaek:


kaidaguerre (commented):
can we add a test for this

PR #4152, opened 2024-02-26
feat: handle ssl password

burahimu:
We figured out that steampipe does not handle private key protected with a passphrase.
This should fix that by adding [sslpassword](https://www.postgresql.org/docs/14/libpq-connect.html#LIBPQ-CONNECT-SSLPASSWORD) and [ssl_passphrase_command](https://www.postgresql.org/docs/14/runtime-config-connection.html#GUC-SSL-PASSPHRASE-COMMAND) options (and [ssl_passphrase_command_supports_reload](https://www.postgresql.org/docs/14/runtime-config-connection.html#GUC-SSL-PASSPHRASE-COMMAND-SUPPORTS-RELOAD))

#### Note

The `server.key` content **must** contains [Proc-Type](https://datatracker.ietf.org/doc/html/rfc1421#section-4.6.1.1) and [DEK-Info](https://datatracker.ietf.org/doc/html/rfc1421#section-4.6.1.3) headers.

It is possible by adding `-traditional` to `openssl` command:

```bash
openssl genrsa -aes256 -out $STEAMPIPE_INSTALL_DIR/db/14.2.0/data/server.key -passout pass:steampipe -traditional 2048
```

### DIFF
--- a/cmd/service.go
+++ b/cmd/service.go
@@ -67,6 +67,7 @@ connection from any Postgres compatible database client.`,
 		AddIntFlag(constants.ArgDatabasePort, constants.DatabaseDefaultPort, "Database service port").
 		AddStringFlag(constants.ArgDatabaseListenAddresses, string(db_local.ListenTypeNetwork), "Accept connections from: `local` (an alias for `localhost` only), `network` (an alias for `*`), or a comma separated list of hosts and/or IP addresses").
 		AddStringFlag(constants.ArgServicePassword, "", "Set the database password for this session").
+		AddStringFlag(constants.ArgDatabaseSSLPassword, "", "Set the database SSL password for the encrypted private key").
 		// default is false and hides the database user password from service start prompt
 		AddBoolFlag(constants.ArgServiceShowPassword, false, "View database password for connecting from another machine").
 		// dashboard server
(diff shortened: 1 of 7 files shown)


e-gineer:
This is a good feature and PR. Some questions:
* Should this focus have CLI arguments etc? It's consistent, but perhaps simpler to start with just the env var for this (edge) case.
* If the password env var is set, would that affect our automatic certificate creation? How do we make that behavior clear?

burahimu:
> Should this focus have CLI arguments etc? It's consistent, but perhaps simpler to start with just the env var for this (edge) case.

Yes it can be set only in env var as a first step.

> If the password env var is set, would that affect our automatic certificate creation? How do we make that behavior clear?

For instance, I believe steampipe will override the `server.key` file with no passphrase.
The certificate renew will be specific to the authority managing the certificates, unless you want steampipe to manage the case?

kaidaguerre:
@burahimu 
> Yes it can be set only in env var as a first step. Would you like this behavior?

Yes, I think for the initial implementation, just an env var would be good

burahimu:
Done ✅

kaidaguerre:
thanks!

PR #4157, opened 2024-02-29
Fix asff removing skip first row condition

sbldevnet:
ASFF exports are skipping the first result, removing the condition will shows all results in the export file.


kaidaguerre:
Many thanks for spotting and fixing this one @sbldevnet

PR #4298, opened 2024-06-05
Fix issue where sps output is not working. Closes #4297

### LINKED ISSUE #4297: `sps` output not working
Steampipe version v0.23.2

```
steampipe query "select 1" --output sps
Error: invalid argument "sps" for "--output" flag: must be 'csv', 'json', 'line', 'snapshot', 'pps', 'table'
```

Steampipe snapshot format should be `sps` and not `pps`. Same behaviour in `check`

pskrbasu:


pkg/version/version.go
```diff
@@ -18,7 +18,7 @@ Also https://www.digitalocean.com/community/tutorials/using-ldflags-to-set-versi
 **/
 
 // The main version number that is being run at the moment.
```
kaidaguerre, line 20: let's update the version in a separate commit which also updates the CHANGELOG, and is has it's own isse

kaidaguerre (changes requested):

PR #4329, opened 2024-07-11
fix: fixed credStore to use Docker when not Turbot GHCR

graza-io:


pdecat (approved):
LGTM! And can confirm this fixes this issue I mentioned in https://github.com/turbot/steampipe/issues/4323#issuecomment-2223003898

PR #4387, opened 2024-09-13
fix: version mismatch for 0.24.1 rel

chenrui333:
seeing the version mismatch while doing the brew release build, https://github.com/Homebrew/homebrew-core/pull/189293. 

cc @pskrbasu


pskrbasu:
Hey @chenrui333, thanks for picking this up, but unfortunately, we would have to release another patch with this fixed. I'll be releasing v0.24.2 shortly.

pskrbasu (approved):

chenrui333:
noted, I was thinking that you probably get away with re-tag, but works for me, thanks!

PR #4564, opened 2025-06-10
exit 1 when a command does not exist

### LINKED ISSUE #4563: Non existant command should return non-zero
**Describe the bug**
Calling steampipe with a command that doesn't exist return 0.

v1.1.4

**To reproduce**
Run this command `steampipe nonexistant`

**Expected behavior**
I expect the exit code to signal failure by returning non-zero

**Additional context**
I use steampipe in CI and I was installing the latest version every CI build. When the `check` command
was removed from steampipe my pipeline did not fail even though the command non longer existed.

frek818:
The change fixes #4563. It exits `1` when `steampipe` is called with a command that does not exist.

### DIFF
--- a/tests/acceptance/test_files/exit_codes.bats
+++ b/tests/acceptance/test_files/exit_codes.bats
@@ -16,7 +16,7 @@ load "$LIB_BATS_SUPPORT/load.bash"
 }
 
 @test "steampipe nonexistant pass with 1 exit code" {
-  # this command should exit one since nonexistant does not exist 
+  # this command should exit one since nonexistent does not exist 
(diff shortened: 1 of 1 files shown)


frek818:
It's currently failing a couple of test [tests](https://github.com/turbot/steampipe/actions/runs/15558029358).

https://github.com/turbot/steampipe/actions/runs/15558029358/job/43807462997#step:13:64

https://github.com/turbot/steampipe/actions/runs/15558029358/job/43807463000#step:13:28

pskrbasu:
Hey @frek818, thank you for the contribution!

PR #4647, opened 2025-08-31
Replaced utils.PidExists with putils.PidExists from pipe-fittings/v2/utils

BelakbirSaid:


pskrbasu:
Hi @BelakbirSaid 

Thank you so much for putting in the effort to make this change. However, this is not the right change. The `PidExists` func in [pipe-fittings](https://github.com/turbot/pipe-fittings/blob/develop/utils/pid_exists.go) is different from the one in [steampipe](https://github.com/turbot/steampipe/blob/develop/pkg/utils/pid_exists.go#L22). 

We have tried making the same change like you did before and we were seeing some errors with the pids not being found resulting in unsuccessful service shutdowns. 

Thanks again!

PR #4720, opened 2025-11-09
Fix #4707: Add nil check in hideRootFlags

### LINKED ISSUE #4707: Bug: Nil pointer dereference in hideRootFlags when called with non-existent flag
## Summary
The `hideRootFlags` function in `cmd/root.go` panics with a nil pointer dereference if called with a flag name that doesn't exist. This could cause the CLI to crash on startup if there's a typo in flag names passed to this function.

## Severity
**HIGH** - Can cause CLI startup crash

## Location
- **File:** `cmd/root.go`
- **Function:** `hideRootFlags()` (line 81)

## Root Cause
`rootCmd.Flag()` returns `nil` for non-existent flags. The code then tries to access `.Hidden` on the nil pointer:

```go
func hideRootFlags(flags ...string) {
	for _, flag := range flags {
		rootCmd.Flag(flag).Hidden = true  // ← nil pointer dereference if flag doesn't exist
	}
}
```

## Reproduction
```go
// This will panic
hideRootFlags("nonexistent-flag-name")

// Error: runtime error: invalid memory address or nil pointer dereference
```

## Test Evidence
Test `TestHideRootFlags_NilPointerBug` in `cmd/root_test.go:12` confirms this bug.

```bash
go test ./cmd -run TestHideRootFlags_NilPointerBug -v
# Output: BUG CONFIRMED: hideRootFlags panics on non-existent flag
```

## Impact
If `InitCmd()` or any other initialization code calls `hideRootFlags` with:
- A typo in flag name
- A flag that was removed but reference not updated
- A flag from a different command

The entire CLI will crash on startup with a panic.

## Suggested Fix
Add a nil check before accessing the flag:

```go
func hideRootFlags(flags ...string) {
	for _, flag := range flags {
		f := rootCmd.Flag(flag)
		if f != nil {  // ← Add nil check
			f.Hidden = true
		}
	}
}
```

## Alternative Fix
Log a warning if flag doesn't exist (for debugging):

```go
func hideRootFlags(flags ...string) {
	for _, flag := range flags {
		f := rootCmd.Flag(flag)
		if f != nil {
			f.Hidden = true
		} else {
			log.Printf("[WARN] Attempted to hide non-existent flag: %s", flag)
		}
	}
}
```

## Found By
Wave 2 Quality-First Testing (Task 4)
Test: `cmd/root_test.go:12`

e-gineer:
## Summary
Fixes #4707 - Nil pointer dereference in `hideRootFlags` when called with non-existent flag

## Problem
The `hideRootFlags` function in `cmd/root.go` was panicking with a nil pointer dereference when called with a flag name that doesn't exist. This occurred because `rootCmd.Flag()` returns `nil` for non-existent flags, and the code attempted to access `.Hidden` on the nil pointer without checking.

This could cause the entire CLI to crash on startup if:
- There's a typo in flag names passed to `hideRootFlags`
- A flag was removed but the reference wasn't updated
- A flag from a different command was mistakenly referenced

## Solution
Added a nil check before accessing the flag's `Hidden` property:

```go
func hideRootFlags(flags ...string) {
	for _, flag := range flags {
		f := rootCmd.Flag(flag)
		if f != nil {
			f.Hidden = true
		}
	}
}
```

## Testing
- The existing test `TestHideRootFlags_NilPointerBug` in `cmd/root_test.go` now passes, confirming the bug is fixed
- All other tests in the `cmd` package continue to pass
- The fix gracefully handles non-existent flags without logging warnings (silent fail is appropriate for this use case)

## Impact
This is a defensive fix that prevents potential CLI startup crashes. The change is backward compatible and doesn't affect normal operation.

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -80,7 +80,10 @@ func InitCmd() {
 
 func hideRootFlags(flags ...string) {
 	for _, flag := range flags {
-		rootCmd.Flag(flag).Hidden = true
+		f := rootCmd.Flag(flag)
+		if f != nil {
+			f.Hidden = true
+		}
 	}
 }
 


e-gineer:
Closing to rework from develop branch with minimal changes

PR #4733, opened 2025-11-09
Fix #4699: Add bounds check to isValidDatabaseName

### LINKED ISSUE #4699: isValidDatabaseName panics on empty string input
## Bug Description

`isValidDatabaseName()` accesses `databaseName[0]` without bounds check, causing a panic when given an empty string.

## Location

- **File:** `pkg/db/db_local/install.go`
- **Function:** `isValidDatabaseName(databaseName string)`

## Severity

**MEDIUM** - Crash if called with empty database name from config or CLI

## Reproduction

```go
isValidDatabaseName("") // PANIC: index out of range [0] with length 0
```

## Likely Implementation

```go
func isValidDatabaseName(databaseName string) bool {
    return databaseName[0] == '_' || (databaseName[0] >= 'a' && databaseName[0] <= 'z')
    // ↑ No bounds check! Panics if len(databaseName) == 0
}
```

## Recommended Fix

Add bounds check before accessing the first character:

```go
func isValidDatabaseName(databaseName string) bool {
    if len(databaseName) == 0 {
        return false
    }
    return databaseName[0] == '_' || (databaseName[0] >= 'a' && databaseName[0] <= 'z')
}
```

## Test

Documented in: `TestIsValidDatabaseName_EmptyString` in `pkg/db/db_local/install_test.go:42-57` (currently skipped)

## Found In

Wave 1.5 test quality assessment - Task 1 (DB & Infrastructure)

## Labels

Priority: MEDIUM, good-first-issue, found-in:wave-1.5

e-gineer:
Fixes #4699

Adds length check before accessing first character to prevent index out of range panic on empty strings.

### DIFF
--- a/pkg/db/db_local/install.go
+++ b/pkg/db/db_local/install.go
@@ -378,6 +378,9 @@ func startServiceForInstall(port int) (*psutils.Process, error) {
 }
 
 func isValidDatabaseName(databaseName string) bool {
+	if len(databaseName) == 0 {
+		return false
+	}
 	return databaseName[0] == '_' || (databaseName[0] >= 'a' && databaseName[0] <= 'z')
 }
 


e-gineer:
Test added with proper two-commit structure:

**Commit 1:** Add test for #4699 (initially expects panic)
- Added `TestIsValidDatabaseName_EmptyString` test case
- Test demonstrates the bug by expecting a panic when calling `isValidDatabaseName("")`
- Test passes because the panic occurs as expected

**Commit 2:** Fix #4699 (applies fix and updates test)
- Added bounds check to `isValidDatabaseName` function to return `false` for empty strings
- Updated test to expect `false` instead of panic
- All tests pass

The fix prevents the panic by checking `len(databaseName) == 0` before attempting to access `databaseName[0]`.

e-gineer:
Updated PR to follow proper TDD structure:

**Commit 1: Test** (251ef0a9)
- Added `TestIsValidDatabaseName_EmptyString` test
- Test expects `false` for empty string input (correct behavior)
- Test FAILS initially because function panics with "index out of range"

**Commit 2: Fix** (b178fbe8)
- Added bounds check to `isValidDatabaseName()` function
- Test file remains UNCHANGED in this commit
- Test now PASSES

This demonstrates true TDD: write failing test first, then implement fix without modifying test.

Verified:
- Test fails on commit 1 (panic)
- Test passes on commit 2 (returns false)
- All other tests in package still pass

PR #4742, opened 2025-11-09
Fix memory leak in DatabaseSession management (issue #3737)

### LINKED ISSUE #3737: Tidy-up `DatabaseSession`
Re-evaluate usage of `db/db_common_DatabaseSession`.

Also there's no cleanup of these objects. We should keep these in sync with the pool.

e-gineer:
## Summary
Fixes #3737 - Memory leak in DatabaseSession management

This PR resolves the memory leak where `DatabaseSession` objects accumulated indefinitely in the sessions map without cleanup during normal operation.

## Changes Made
- **Added BeforeClose callback** in `pkg/db/db_client/db_client_connect.go` to automatically clean up session map entries when connections are closed by pgx
- **Updated comment** in `pkg/db/db_client/db_client.go` to reflect that sessions are now automatically cleaned up
- **Enhanced TestSessionMapLeak** to verify cleanup behavior works correctly
- **Added TestSessionCleanupOnConnectionClose** - new comprehensive test that validates the fix by simulating connection cleanup

## Technical Details
The `sessions` map in `DbClient` tracks database connections keyed by backend process IDs. Previously, when connections were dropped by pgx, the corresponding session entries remained in the map indefinitely, causing memory to accumulate over time in long-running services.

The fix uses pgxpool's `BeforeClose` callback, which is invoked automatically whenever a connection is closed. This ensures session map entries are removed promptly rather than accumulating until the entire client is closed.

## Test Plan
✅ All existing tests pass
✅ `TestSessionMapLeak` - verifies sessions can be cleaned up during operation
✅ `TestSessionCleanupOnConnectionClose` - validates the BeforeClose callback mechanism
✅ `TestConcurrentSessionAccess` - ensures thread-safety with cleanup

Run tests with:
```bash
go test ./pkg/db/db_client/ -v
```

## Impact
- Prevents memory accumulation in long-running services with connection churn
- Minimal change - only adds cleanup callback, no changes to existing logic
- Thread-safe using existing sessionsMutex

### DIFF
--- a/pkg/db/db_client/db_client.go
+++ b/pkg/db/db_client/db_client.go
@@ -42,8 +42,9 @@ type DbClient struct {
 
 	// map of database sessions, keyed to the backend_pid in postgres
 	// used to update session search path where necessary
-	// sessions are automatically cleaned up via BeforeClose callback when connections are dropped
-	// https://github.com/turbot/steampipe/issues/3737
+	// Session lifecycle: entries are added when connections are established and automatically
+	// removed via a pgxpool BeforeClose callback when connections are closed by the pool.
+	// This prevents memory accumulation from stale connection entries (see issue #3737)
 	sessions map[uint32]*db_common.DatabaseSession
 
 	// allows locked access to the 'sessions' map


e-gineer:
Thanks for the feedback! I've improved the comment to be more descriptive. The updated comment now explains:

- **Session lifecycle**: When entries are added and removed
- **Mechanism**: Specifically mentions pgxpool BeforeClose callback
- **Benefit**: Explains how this prevents memory accumulation

The new comment provides better context for future maintainers about how the session map is managed.

PR #4826, opened 2025-11-11
Fix #4781: RunBatchSession blocks forever if initData.Loaded never closes

### LINKED ISSUE #4781: RunBatchSession blocks forever if initData.Loaded never closes
## Description
The `RunBatchSession` function unconditionally blocks on `<-initData.Loaded` at line 58 without checking context cancellation or setting a timeout. If the initialization process that's supposed to close the `Loaded` channel hangs, panics, or never completes, `RunBatchSession` will block forever.

This is particularly problematic because:
1. The function doesn't respect context cancellation before blocking
2. There's no timeout mechanism
3. The goroutine that's supposed to close `Loaded` may never run or may crash

## Severity
**HIGH** - Application hangs indefinitely

## Reproduction
1. Create an InitData where the Loaded channel never closes
2. Call `RunBatchSession(ctx, initData)`
3. Observe that the function hangs indefinitely, even if context is cancelled

## Expected Behavior
Should respect context cancellation:
```go
func RunBatchSession(ctx context.Context, initData *query.InitData) (int, error) {
    contexthelpers.StartCancelHandler(initData.Cancel)

    // Wait for init with context support
    select {
    case <-initData.Loaded:
        // Continue
    case <-ctx.Done():
        return 0, ctx.Err()
    }

    if err := initData.Result.Error; err != nil {
        return 0, err
    }
    // ... rest of function
}
```

## Test Reference
Test: `TestRunBatchSession_LoadedTimeout` in `pkg/query/queryexecute/execute_test.go:75` (currently skipped)

## Suggested Fix
Use a select statement to check both the Loaded channel and context cancellation.

## Related Code
- `pkg/query/queryexecute/execute.go:58` - Blocking receive without context check

e-gineer:
## Summary
- Add test demonstrating bug where RunBatchSession blocks forever
- Fix RunBatchSession to respect context cancellation

## Test plan
- Test passes showing RunBatchSession now respects context cancellation
- Test previously would hang forever demonstrating the bug

## Changes
1. **Commit 1**: Add test demonstrating the bug (initially skipped)
   - Creates TestRunBatchSession_LoadedTimeout that shows RunBatchSession blocks forever if initData.Loaded never closes
   
2. **Commit 2**: Fix the bug and unskip test
   - Changes line 58 from blocking `<-initData.Loaded` to `select` statement with `ctx.Done()` check
   - Unskips the test which now passes

Fixes #4781 - closes #4781

### DIFF
--- a/pkg/query/queryexecute/execute.go
+++ b/pkg/query/queryexecute/execute.go
@@ -58,8 +58,14 @@ func RunBatchSession(ctx context.Context, initData *query.InitData) (int, error)
 	// NOTE: use the initData Cancel function to ensure any initialisation is cancelled if needed
 	contexthelpers.StartCancelHandler(initData.Cancel)
 
-	// wait for init
-	<-initData.Loaded
+	// wait for init, respecting context cancellation
+	select {
+	case <-initData.Loaded:
+		// initialization complete, continue
+	case <-ctx.Done():
+		// context cancelled before initialization completed
+		return 0, ctx.Err()
+	}
 
 	if err := initData.Result.Error; err != nil {
 		return 0, err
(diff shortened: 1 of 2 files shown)


e-gineer:
This PR does not show the test as failing because the bug causes the test to hang forever. Instead, the test is demonstrated and skipped. The fix removes the skip on the test.

This is ready to merge!

PR #4832, opened 2025-11-11
Race condition between StreamRow() and Close() closes #4790

### LINKED ISSUE #4790: Race condition between StreamRow() and Close()
## Description

## Severity

## Reproduction

## Expected Behavior

## Test Reference

## Suggested Fix
```go
type Result struct {
    closeOnce sync.Once
    closed    atomic.Bool
}

    if !r.closed.Load() {
        r.Result.StreamRow(row)
    }
}
```

## Related Code

e-gineer:
## Summary
Fixes a race condition in `pkg/query/queryresult/result.go` where `StreamRow()` and `StreamError()` can be called concurrently with `Close()`, causing a "send on closed channel" panic when Close() closes the RowChan while StreamRow() or StreamError() tries to send.

## Changes
- **Commit 1**: Added test demonstrating the race condition - test fails with `-race` flag showing concurrent access to the channel
- **Commit 2**: Fixed the race by adding RWMutex protection and a closed flag to the Result wrapper
  - StreamRow/StreamError acquire read lock and check closed flag before sending
  - Close acquires write lock and sets closed flag before closing channel
  - This prevents any sends after the channel is closed without blocking concurrent StreamRow calls

## Test plan
- Commit 1 shows the test failing with race detector, demonstrating the bug
- Commit 2 shows the test passing with the fix applied
- Run `go test -race ./pkg/query/queryresult` to verify no race conditions

### DIFF
--- a/pkg/query/queryresult/result.go
+++ b/pkg/query/queryresult/result.go
@@ -10,6 +10,8 @@ import (
 type Result struct {
 	*queryresult.Result[TimingResultStream]
 	closeOnce sync.Once
+	mu        sync.RWMutex
+	closed    bool
 }
 
 func NewResult(cols []*queryresult.ColumnDef) *Result {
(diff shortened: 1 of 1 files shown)


e-gineer:
Closing this PR in favor of the rebased version.

**Superseded by**: PR #4876 which includes the same fix rebased onto the latest develop branch with CI properly showing fail→pass pattern.

The fix is still valid and has been carried forward to the new PR.

PR #4837, opened 2025-11-11
Fix race condition on customSearchPath slice closes #4792

### LINKED ISSUE #4792: Potential race on customSearchPath slice
## Description
The `customSearchPath []string` field is accessed from multiple goroutines without synchronization. While reads are generally safe, modifications to the slice (append operations) can cause races.

## Severity
**MEDIUM** - Data race on slice

## Test Reference
Test: `TestDbClient_SessionSearchPathUpdatesThreadSafe` in `pkg/db/db_client/db_client_session_test.go:178` (skipped)

## Suggested Fix
Either make customSearchPath immutable after initialization or protect all access with a mutex.

## Related Code
- `pkg/db/db_client/db_client.go:54`

e-gineer:
## Summary
Fixed a data race on the `customSearchPath` and `searchPathPrefix` slices in `DbClient`. These fields were being accessed concurrently from multiple goroutines without synchronization, which could cause race conditions when the race detector is enabled.

The fix adds a dedicated `searchPathMutex` to protect all reads and writes to these fields.

## Changes
- **Commit 1**: Added test demonstrating the race condition
  - Created `TestDbClient_SessionSearchPathUpdatesThreadSafe` which spawns 300 concurrent goroutines that read and write to the search path fields
  - Test fails with `-race` flag before the fix, demonstrating the bug
  
- **Commit 2**: Implemented the fix
  - Added `searchPathMutex` field to `DbClient` struct
  - Protected all `customSearchPath` and `searchPathPrefix` reads/writes with the mutex in:
    - `SetRequiredSessionSearchPath` (writes)
    - `GetRequiredSessionSearchPath` (reads)
    - `GetCustomSearchPath` (reads)
  - Fixed logging to use already-fetched value instead of direct field access
  - Updated test to initialize the mutex for proper testing

## Test plan
- [x] Checkout commit 1, run `go test -race -run TestDbClient_SessionSearchPathUpdatesThreadSafe ./pkg/db/db_client` - should FAIL with race detector warnings
- [x] Checkout commit 2, run `go test -race -run TestDbClient_SessionSearchPathUpdatesThreadSafe ./pkg/db/db_client` - should PASS
- [ ] CI tests pass

### DIFF
--- a/pkg/db/db_client/db_client.go
+++ b/pkg/db/db_client/db_client.go
@@ -54,6 +54,8 @@ type DbClient struct {
 	// if a custom search path or a prefix is used, store it here
 	customSearchPath []string
 	searchPathPrefix []string
+	// allows locked access to customSearchPath and searchPathPrefix
+	searchPathMutex *sync.Mutex
 	// the default user search path
 	userSearchPath []string
 	// disable timing - set whilst in process of querying the timing
@@ -71,6 +73,7 @@ func NewDbClient(ctx context.Context, connectionString string, opts ...ClientOpt
 		parallelSessionInitLock: semaphore.NewWeighted(constants.MaxParallelClientInits),
 		sessions:                make(map[uint32]*db_common.DatabaseSession),
 		sessionsMutex:           &sync.Mutex{},
+		searchPathMutex:         &sync.Mutex{},
 		connectionString:        connectionString,
 	}
 
--- a/pkg/db/db_client/db_client_search_path.go
+++ b/pkg/db/db_client/db_client_search_path.go
@@ -29,9 +29,6 @@ func (c *DbClient) SetRequiredSessionSearchPath(ctx context.Context) error {
 	// default required path to user search path
 	requiredSearchPath := c.userSearchPath
 
-	// store custom search path and search path prefix
-	c.searchPathPrefix = searchPathPrefix
-
 	// if a search path was passed, use that
 	if len(configuredSearchPath) > 0 {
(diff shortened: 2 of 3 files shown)


e-gineer:
## PR Status Update

This PR has been rebased onto the latest `develop` branch and follows the 2-commit pattern (test then fix).

### Why the test commit passed CI

The test commit (5660e84b) passed CI even though it demonstrates a race condition. This is because:

1. **Race conditions require the `-race` flag to detect**: The Go race detector must be explicitly enabled with `go test -race`
2. **CI doesn't run tests with `-race`**: The current CI configuration runs `go test` without the `-race` flag
3. **Local verification confirms the race**: When run locally with `-race`, commit 1 fails with multiple data race warnings:
   ```
   ==================
   WARNING: DATA RACE
   Write at 0x00c00078e2e0 by goroutine 240:
     github.com/turbot/steampipe/v2/pkg/db/db_client.(*DbClient).SetRequiredSessionSearchPath()
         /tmp/steampipe-pr-4837/pkg/db/db_client/db_client_search_path.go:33
   ```

### Verification

Both commits have been tested locally with the race detector:
- **Commit 1** (5660e84b - test only): `go test -race` → **FAIL** with data race warnings ✅
- **Commit 2** (f2328dc8 - with fix): `go test -race` → **PASS** with no race conditions ✅

The fix properly adds `searchPathMutex` protection to eliminate the race condition on `customSearchPath` and `searchPathPrefix` slices.

PR #4847, opened 2025-11-11
WrapResult nil input handling closes #4804

### LINKED ISSUE #4804: WrapResult may not initialize closeOnce for nil input
## Description
The `WrapResult()` function accepts a pointer but doesn't validate if it's nil. Calling any method on the wrapped nil result will cause a nil pointer dereference.

## Severity
**MEDIUM** - Nil pointer dereference

## Test Reference
Test: `TestWrapResult_NilResult` in `pkg/query/queryresult/result_test.go:214` (skipped)

## Suggested Fix
```go
func WrapResult(r *queryresult.Result[TimingResultStream]) *Result {
    if r == nil {
        return nil
    }
    return &Result{Result: r}
}
```

## Related Code
- `pkg/query/queryresult/result.go:29-32`

e-gineer:
## Summary
WrapResult did not handle nil input gracefully, returning a wrapper around nil instead of nil itself. This could cause unexpected behavior when calling methods on the result.

## Changes
- Commit 1: Unskipped test demonstrating the bug
- Commit 2: Added nil check to return nil when input is nil

## Test Results
- Before fix: Test fails - WrapResult(nil) returns a non-nil wrapper
- After fix: Test passes - WrapResult(nil) returns nil

## Verification
```bash
# Commit 1 (test only)
go test -v -run TestWrapResult_NilResult ./pkg/query/queryresult
# FAIL: Expected nil, but got wrapper around nil

# Commit 2 (with fix)
go test -v -run TestWrapResult_NilResult ./pkg/query/queryresult
# PASS
```

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/query/queryresult/result.go
+++ b/pkg/query/queryresult/result.go
@@ -27,6 +27,9 @@ func (r *Result) Close() {
 
 // WrapResult wraps a pipe-fittings Result with our wrapper that has idempotent Close
 func WrapResult(r *queryresult.Result[TimingResultStream]) *Result {
+	if r == nil {
+		return nil
+	}
 	return &Result{
 		Result: r,
 	}


e-gineer:
This PR has been replaced by #4875 which shows the proper fail → pass CI pattern.

The commits are identical, but #4875 demonstrates:

PR #4848, opened 2025-11-11
initialisationComplete flag synchronized closes #4803

### LINKED ISSUE #4803: initialisationComplete flag not synchronized
## Description
Boolean flag read/written by multiple goroutines (main, init, executor) without synchronization.

## Severity
**MEDIUM** - Data race

## Impact
Query execution may proceed before initialization complete, or spinner shows incorrectly.

## Suggested Fix
Use `atomic.Bool` for the flag.

## Related Code
- `pkg/interactive/interactive_client.go:61`
- `pkg/interactive/interactive_client_init.go:19,129`

e-gineer:
## Summary
Fixes a data race condition where the `initialisationComplete` boolean flag was being accessed concurrently by multiple goroutines without synchronization. The flag is written by the initialization goroutine and read by the query executor and notification handler.

## Changes
- **Commit 1**: Added test demonstrating the race condition
  - Test simulates concurrent reads and writes to the flag
  - Fails with `-race` flag before the fix
- **Commit 2**: Implemented fix using `atomic.Bool`
  - Changed `initialisationComplete` from `bool` to `atomic.Bool`
  - Updated all writes to use `.Store()`
  - Updated all reads to use `.Load()`

## Test Results
- **Before fix**: Race detector reports data race in multiple locations
- **After fix**: All tests pass with `-race` flag

## Verification
```bash
# Commit 1 (test only)
go test -race -v -run TestInitialisationComplete_RaceCondition ./pkg/interactive
# FAIL: race detected during execution of test

# Commit 2 (with fix)
go test -race -v -run TestInitialisationComplete_RaceCondition ./pkg/interactive
# PASS
```

The two-phase push ensures CI runs separately on each commit, demonstrating:
1. First CI run: Test fails with race detection (proving the bug exists)
2. Second CI run: Test passes (proving the fix works)

### DIFF
--- a/pkg/interactive/interactive_client.go
+++ b/pkg/interactive/interactive_client.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/alecthomas/chroma/formatters"
@@ -58,7 +59,7 @@ type InteractiveClient struct {
 	// channel used internally to pass the initialisation result
 	initResultChan chan *db_common.InitResult
 	// flag set when initialisation is complete (with or without errors)
-	initialisationComplete bool
+	initialisationComplete atomic.Bool
 	afterClose             AfterPromptCloseAction
 	// lock while execution is occurring to avoid errors/warnings being shown
 	executionLock sync.Mutex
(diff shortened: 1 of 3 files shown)


e-gineer:
Closing this PR due to merge conflicts with #4864. The fix has been rebased and is now available in #4886.

PR #4851, opened 2025-11-11
EnsureDBInstalled TOCTOU race condition closes #4818

### LINKED ISSUE #4818: EnsureDBInstalled race condition (TOCTOU)
## Description
While `EnsureDBInstalled()` uses a mutex (`ensureMux`) to protect the installation process, the initial check `IsDBInstalled()` at line 56 happens outside the mutex lock. This creates a TOCTOU (Time-of-Check-Time-of-Use) race condition where two concurrent calls could both see the DB as not installed, then both proceed with installation (though the mutex will serialize the actual install).

## Severity
**MEDIUM** - Race condition

## Test Reference
Test: `TestEnsureDBInstalled_Concurrent` in `pkg/db/db_local/install_test.go:167` (skipped)

## Suggested Fix
Move the `IsDBInstalled()` check inside the mutex lock.

## Related Code
- `pkg/db/db_local/install.go:39-54`

e-gineer:
## Summary
Fixes a Time-of-Check-Time-of-Use (TOCTOU) race condition in `EnsureDBInstalled()` where concurrent callers could experience race conditions when checking if the database is installed and then acting on that information.

## Changes
- **Commit 1**: Added test demonstrating the TOCTOU race pattern with concurrent calls to `EnsureDBInstalled()`
- **Commit 2**: Implemented double-checked locking pattern to fix the race:
  1. Fast-path check before acquiring mutex (reduces contention in common case)
  2. Acquire mutex only if DB is not installed
  3. Double-check after acquiring mutex (handles concurrent installation attempts)

## Test Results
The test `TestEnsureDBInstalled_Concurrent` now passes with the `-race` flag, demonstrating thread-safe concurrent access to `EnsureDBInstalled()`.

**Before fix**: Potential race condition where multiple goroutines could check `IsDBInstalled()` outside of mutex protection

**After fix**: All checks are properly synchronized using double-checked locking pattern

## Verification
```bash
# Run the test with race detector
go test -v -race -run TestEnsureDBInstalled_Concurrent ./pkg/db/db_local
# PASS
```

### DIFF
--- a/pkg/db/db_local/install.go
+++ b/pkg/db/db_local/install.go
@@ -40,6 +40,18 @@ If you need to restore the contents of your public schema, please open an issue 
 func EnsureDBInstalled(ctx context.Context) (err error) {
 	putils.LogTime("db_local.EnsureDBInstalled start")
 
+	// Fix #4818: Fast-path check before acquiring mutex to avoid unnecessary contention
+	// when DB is already installed (common case)
+	if IsDBInstalled() {
+		// Still need to prepare the DB even if already installed
+		ensureMux.Lock()
+		defer ensureMux.Unlock()
+		err := prepareDb(ctx)
+		putils.LogTime("db_local.EnsureDBInstalled end")
+		return err
+	}
+
+	// DB not installed - acquire mutex for installation
 	ensureMux.Lock()
 
 	doneChan := make(chan bool, 1)
@@ -53,6 +65,7 @@ func EnsureDBInstalled(ctx context.Context) (err error) {
 		close(doneChan)
 	}()
 
+	// Double-check after acquiring mutex - another goroutine may have installed it
 	if IsDBInstalled() {
 		// check if the FDW need updating, and init the db if required
 		err := prepareDb(ctx)


e-gineer:
Closing this PR as the reported bug does not exist in the current codebase.

**Investigation findings:**

The issue claimed there was a TOCTOU (time-of-check to time-of-use) race condition in `EnsureDBInstalled`, where multiple goroutines could check `IsDBInstalled()` before the mutex was acquired.

However, examining the current code in `pkg/db/db_local/install.go` (lines 42-56), the mutex is acquired **BEFORE** the check:

```go
ensureMux.Lock()
defer func() { 
    ensureMux.Unlock()
}()

if IsDBInstalled() {  // Check happens INSIDE the mutex lock
    err := prepareDb(ctx)
    return err
}
```

This is the correct pattern - the `IsDBInstalled()` check is protected by the mutex, so no TOCTOU race exists.

**Conclusion:**
The bug is already fixed (or never existed) in the current develop branch. No code changes needed.

PR #4853, opened 2025-11-11
TEST: Two-phase CI verification (DO NOT MERGE)

e-gineer:
Testing if creating PR after commit 1, then pushing commit 2, gives us two separate CI runs.


e-gineer:
Test successful! Two separate CI runs confirmed. The key is to create the PR after pushing commit 1, then push commit 2.

PR #4861, opened 2025-11-11
Fix logic error in IdentifyMissingComments closes #4814

### LINKED ISSUE #4814: Logic error in IdentifyMissingComments
## Description
The `IdentifyMissingComments()` function has a logic error on line 426. Uses OR (`||`) when it should use AND (`&&`). Current logic: "if NOT updating OR deleting" means connections being deleted are still added to MissingComments, contradicting intended behavior.

## Severity
**LOW** - Wasted work but doesn't cause incorrect behavior

## Reproduction
1. Create ConnectionUpdates with a connection marked for deletion
2. Set CommentsSet to false for that connection
3. Call IdentifyMissingComments()
4. Observe that the deleted connection is added to MissingComments

## Expected Behavior
```go
if \!updating && \!deleting {
    log.Printf("[TRACE] connection %s comments not set, marking as missing", name)
    u.MissingComments[name] = state
}
```

## Test Reference
Test: `TestConnectionUpdates_IdentifyMissingComments` in `pkg/steampipeconfig/connection_updates_test.go:125`

## Related Code
- `pkg/steampipeconfig/connection_updates.go:426`

e-gineer:
## Summary
Fixes a logic error in the `IdentifyMissingComments()` function where OR (`||`) was used instead of AND (`&&`), causing connections being deleted to be incorrectly added to MissingComments.

## Bug Details
- **Location**: `pkg/steampipeconfig/connection_updates.go:426`
- **Current buggy logic**: `if !updating || deleting`
- **Correct logic**: `if !updating && !deleting`
- **Impact**: Connections being deleted were incorrectly marked as missing comments, resulting in wasted work

## Changes
- Commit 1: Added test demonstrating the bug
- Commit 2: Implementing fix (coming in next push)

## Test Results
- **Commit 1 (test only)**: Test should FAIL, demonstrating the bug exists
- **Commit 2 (with fix)**: Test should PASS, proving the fix works

## Verification
```bash
# Before fix (commit 1)
go test -v -run TestConnectionUpdates_IdentifyMissingComments ./pkg/steampipeconfig
# FAIL: connection_being_deleted_should_NOT_be_in_MissingComments
#       expected shouldBeMissing=false, got inMissingComments=true

# After fix (commit 2)
go test -v -run TestConnectionUpdates_IdentifyMissingComments ./pkg/steampipeconfig
# PASS
```

### DIFF
--- a/pkg/steampipeconfig/connection_updates.go
+++ b/pkg/steampipeconfig/connection_updates.go
@@ -423,7 +423,7 @@ func (u *ConnectionUpdates) IdentifyMissingComments() {
 			if !currentState.CommentsSet {
 				_, updating := u.Update[name]
 				_, deleting := u.Delete[name]
-				if !updating || deleting {
+				if !updating && !deleting {
 					log.Printf("[TRACE] connection %s comments not set, marking as missing", name)
 					u.MissingComments[name] = state
 				}


e-gineer:
This PR has been replaced by #4872 which shows the proper fail → pass CI pattern.

The commits are identical, but #4872 demonstrates:

PR #4863, opened 2025-11-11
Certificate rotation with backup closes #4819

e-gineer:
## Summary
In `removeExpiringSelfIssuedCertificates()`, when certificates are removed for regeneration, the old certificates are deleted without backup. If regeneration fails, users lose their certificates with no way to recover. This PR adds timestamped backups before certificate removal, similar to how database backups are retained.

## Changes
- Commit 1: Unskipped test demonstrating the bug
- Commit 2: Implemented fix to create timestamped backups before certificate rotation

## Implementation
The fix adds a `backupCertificateFile()` helper function that:
- Creates timestamped backups in the `backups` directory
- Uses format: `server.crt-2025-11-12-14-30-45`
- Backs up certificates before removal in both `removeServerCertificate()` and `removeAllCertificates()`
- Follows the same backup pattern used for database backups

### Commit 1 (Test only - demonstrates bug)
```bash
go test -v -run TestRemoveExpiringSelfIssuedCertificates ./pkg/db/db_local
# FAIL: BUG #4819: Certificate backup was not created before removal
```

### Commit 2 (With fix - test passes)
```bash
go test -v -run TestRemoveExpiringSelfIssuedCertificates ./pkg/db/db_local
# PASS - Backups are created before certificate removal
```

## CI Verification
- First CI run (commit 1): Failed - demonstrating the bug exists
- Second CI run (commit 2): Should pass - proving the fix works

Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #4870, opened 2025-11-13
Certificate rotation with backup closes #4819

### LINKED ISSUE #4819: Certificate rotation without backup
## Description
In `removeExpiringSelfIssuedCertificates()`, when certificates are removed for regeneration, the old certificates are deleted without backup. If regeneration fails, users lose their certificates with no way to recover.

## Severity
**LOW** - Data loss potential

## Test Reference
Test: `TestRemoveExpiringSelfIssuedCertificates` in `pkg/db/db_local/ssl_test.go:86` (skipped)

## Suggested Fix
Create timestamped backups of old certificates before removal, similar to how database backups are retained.

## Related Code
- `pkg/db/db_local/ssl.go:33-63`

e-gineer:
This PR demonstrates and fixes the issue.

**Two-commit pattern**:
- Commit 1: Test demonstrating the bug (this commit - **should FAIL CI**)
- Commit 2: Fix implementation (will be pushed after CI confirms failure)

This is a re-trigger of PR #4863 to show the proper fail → pass CI pattern.

Closes the same issue as #4863

### DIFF
--- a/pkg/db/db_local/ssl.go
+++ b/pkg/db/db_local/ssl.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"math/big"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


e-gineer:
Closing this PR after further analysis. While the fix follows good backup practices, the actual risk is minimal:

**Why the risk is low:**
1. These are self-issued certificates that can be regenerated at any time using standard crypto libraries
2. The root private key persists across rotations (not deleted), ensuring continuity
3. Regeneration failures (disk full, permissions) would also prevent writing backups
4. No actual user data or irreplaceable material is at risk

**Assessment:**
- The certificates are designed to be ephemeral and regeneratable
- The backup adds complexity for minimal practical benefit
- Better to focus on more impactful fixes

Closing as low priority. Issue #4819 will also be closed.

Issue #4805, opened 2025-11-11
Concurrent read and close may race


e-gineer:
## Description
When one goroutine is reading from `result.RowChan` while another calls `Close()`, there's a potential race condition. The `sync.Once` protects Close() itself but doesn't coordinate with active readers.

## Severity
**MEDIUM** - Race condition

## Test Reference
Test: `TestResult_ConcurrentReadAndClose` in `pkg/query/queryresult/result_test.go:253` (skipped)

## Status
Needs investigation to determine if current behavior is acceptable or if explicit coordination is required.

## Related Code
- `pkg/query/queryresult/result.go:22-26`

PR #4869, opened 2025-11-13
Fix getQueryInfo() 'from ' detection closes #4810

e-gineer:
This PR demonstrates and fixes issue #4810 where getQueryInfo() fails to detect 'from ' (with trailing space) correctly.

**Two-commit pattern**:
- Commit 1: Test demonstrating the bug (this commit - **should FAIL CI**)
- Commit 2: Fix implementation (will be pushed after CI confirms failure)

This PR uses a new branch name to test if CI triggers properly.

PR #4876, opened 2025-11-14
Race condition between StreamRow() and Close() closes #4790

### LINKED ISSUE #4790: Race condition between StreamRow() and Close()

e-gineer:
This PR demonstrates and fixes the issue.

**Two-commit pattern**:
- Commit 1: Test demonstrating the bug (**should FAIL CI**)
- Commit 2: Fix implementation (**should PASS CI**)

This is a re-trigger of PR #4832 to show the proper fail → pass CI pattern.

Closes the same issue as #4832

### DIFF
--- a/pkg/query/queryresult/result.go
+++ b/pkg/query/queryresult/result.go
@@ -31,13 +31,21 @@ func (r *Result) Close() {
 	})
 }
 
-// StreamRow wraps the underlying StreamRow with synchronization
-func (r *Result) StreamRow(row []interface{}) {
+// StreamRow safely sends a row to the RowChan, checking if it's closed first
+func (r *Result) StreamRow(rowResult []interface{}) {
 	r.mu.RLock()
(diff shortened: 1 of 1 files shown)


e-gineer:
Closing as duplicate - bug #4790 was already fixed by PR #4869 (which fixed bug #4805 with the same root cause: race condition between StreamRow() and Close()).

The mutex protection added in #4869 already prevents the 'send on closed channel' panic that this PR was intended to fix.

The only additional change in this PR is the StreamError() wrapper, which is a minor enhancement that's not critical since StreamError is less commonly used than StreamRow.

PR #4737, opened 2025-11-09
Fix #4716: Add synchronization to autoCompleteSuggestions.sort()

e-gineer:
Fixes #4716

Adds mutex protection to prevent data race during concurrent sort() calls.

Issue #4716, opened 2025-11-09
[Bug] Data race in autoCompleteSuggestions.sort() during concurrent access


e-gineer:
## Bug Description

The `sort()` method in `pkg/interactive/autocomplete_suggestions.go` has a data race when called concurrently from multiple goroutines. This can lead to data corruption or crashes during autocomplete initialization.

## Location

**File:** `pkg/interactive/autocomplete_suggestions.go`  
**Lines:** 23-39  
**Function:** `autoCompleteSuggestions.sort()`

## Issue

The `sort()` method modifies slices in place without any synchronization mechanism. When multiple goroutines call `sort()` (or `initialiseSuggestions()` which calls `sort()`), they race to read and write the same memory locations.

```go
func (s *autoCompleteSuggestions) sort() {
    sortSuggestions := func(suggestions []prompt.Suggest) {
        sort.Slice(suggestions, func(i, j int) bool {
            return suggestions[i].Text < suggestions[j].Text
            // ^^^ DATA RACE: Concurrent reads/writes to suggestions slice
        })
    }

    sortSuggestions(s.schemas)
    sortSuggestions(s.unqualifiedTables)
    // ... more sorts
}
```

## Impact

- **Severity:** Medium
- **Symptoms:** Data corruption, crashes, inconsistent autocomplete suggestions
- **Conditions:** Occurs when autocomplete is initialized from multiple goroutines simultaneously
- **Race Detector Output:**
  ```
  WARNING: DATA RACE
  Read at 0x00c000820d20 by goroutine 27
  Previous write at 0x00c000820d20 by goroutine 29
  ```

## Steps to Reproduce

1. Run the test with race detector:
   ```bash
   go test -race ./pkg/interactive/ -run TestAutocompleteSuggestions_ConcurrentSort
   ```
2. Race detector reports data race
3. Or, call `sort()` from multiple goroutines:
   ```go
   suggestions := newAutocompleteSuggestions()
   go suggestions.sort()
   go suggestions.sort()
   go suggestions.sort()
   ```

## Detection

Found via concurrent access testing with Go's race detector:
- Test: `TestAutocompleteSuggestions_ConcurrentSort`
- File: `pkg/interactive/interactive_client_autocomplete_extended_test.go`
- Run with: `go test -race`

### Option 1: Add Mutex Protection
```go
type autoCompleteSuggestions struct {
    mu                  sync.RWMutex
    schemas             []prompt.Suggest
[...]

PR #4903, opened 2025-11-16
Fix #4711: Race condition in autoCompleteSuggestions.sort() method

### LINKED ISSUE #4711: Bug: Race condition in autoCompleteSuggestions.sort() method
## Bug Description
The `sort()` method in `autoCompleteSuggestions` has a race condition when called concurrently. The method modifies shared slice data without synchronization, leading to data races detected by Go's race detector.

## Location
**File:** `pkg/interactive/autocomplete_suggestions.go`  
**Line:** 23  
**Function:** `sort()`

## Impact
- **Severity:** Medium
- **Type:** Race condition / data race
- Concurrent calls to `sort()` modify underlying slice arrays
- Detected by `-race` flag
- Could lead to inconsistent or corrupted suggestion ordering

## Race Detector Output
```
==================
WARNING: DATA RACE
Read at 0x... by goroutine X:
  sort.StringSlice.Less()
  pkg/interactive/autocomplete_suggestions.go:26
  
Write at 0x... by goroutine Y:
  sort.StringSlice.Swap()
  pkg/interactive/autocomplete_suggestions.go:26
```

## Code
```go
func (s autoCompleteSuggestions) sort() {
    sortSuggestions := func(s []prompt.Suggest) {
        sort.Slice(s, func(i, j int) bool {
            return s[i].Text < s[j].Text
        })
    }

    sortSuggestions(s.schemas)
    sortSuggestions(s.unqualifiedTables)
    sortSuggestions(s.unqualifiedQueries)
    for _, tables := range s.tablesBySchema {
        sortSuggestions(tables)  // Race here
    }
    for _, queries := range s.queriesByMod {
        sortSuggestions(queries)  // Race here
    }
}
```

## How It Was Found
- **Test:** Wave 2 Task 6 comprehensive testing
- **Method:** Concurrent access testing with race detector
- Test file: `pkg/interactive/interactive_client_autocomplete_extended_test.go`
- Test: `TestAutocompleteSuggestions_ConcurrentSort`
- Command: `go test -race ./pkg/interactive/...`

## Suggested Fix
Add synchronization to prevent concurrent modification:

**Option 1: Mutex**
```go
type autoCompleteSuggestions struct {
    mu                 sync.Mutex  // Add mutex
    schemas            []prompt.Suggest
    // ... rest of fields
}

func (s *autoCompleteSuggestions) sort() {
    s.mu.Lock()
    defer s.mu.Unlock()
    // ... existing sort logic
}
```

**Option 2: Only sort once after initialization**
Ensure sort is only called during initialization, not concurrently during suggestions retrieval.

## Occurrences
The sort method is called from:
1. `pkg/interactive/interactive_client_autocomplete.go:37` - `initialiseSuggestions()`
2. Multiple test files

Review: Is sort() ever called concurrently in production? If not, the fix may be to ensure it's only called during initialization.

## Status
⚠️ **FOUND BUT NOT FIXED** - Needs design decision on synchronization approach

## Related
Part of Wave 2 Task 6 (Interactive Client Tests) bug hunting effort.
Found 4 bugs total following Wave 1.5 quality standards.

e-gineer:
## Summary
Fixed a race condition in the `autoCompleteSuggestions.sort()` method that occurred when the method was called concurrently, leading to data races on the underlying slice arrays during sorting operations.

## Root Cause
The `sort()` method in `autoCompleteSuggestions` modified shared slice data without any synchronization. When called concurrently from multiple goroutines, this caused data races as detected by Go's race detector. The method performed in-place sorting on multiple slices (schemas, tables, queries) and map values, allowing multiple goroutines to simultaneously read and write to the same memory locations.

## Solution
Added a `sync.Mutex` to the `autoCompleteSuggestions` struct to synchronize access to the sort method. The sort() method now:
- Changed receiver from value to pointer (`*autoCompleteSuggestions`)
- Acquires a lock before performing any sorting operations
- Releases the lock after all sorting is complete

This ensures that only one goroutine can execute the sort operation at a time, preventing concurrent modification of the underlying data structures.

## Testing
- Added test `TestAutocompleteSuggestions_ConcurrentSort` demonstrating the race condition
- Verified test fails with race detector before fix
- Verified test passes with race detector after fix
- All existing tests continue to pass with `-race` flag

Fixes #4711

### DIFF
--- a/pkg/interactive/autocomplete_suggestions.go
+++ b/pkg/interactive/autocomplete_suggestions.go
@@ -1,8 +1,10 @@
 package interactive
 
 import (
-	"github.com/c-bata/go-prompt"
 	"sort"
+	"sync"
+
+	"github.com/c-bata/go-prompt"
 )
 
 const (
@@ -15,6 +17,7 @@ const (
 )
 
 type autoCompleteSuggestions struct {
+	mu                 sync.Mutex
 	schemas            []prompt.Suggest
 	unqualifiedTables  []prompt.Suggest
 	unqualifiedQueries []prompt.Suggest
@@ -72,7 +75,10 @@ func (s *autoCompleteSuggestions) setQueriesForMod(modName string, queries []pro
 	s.queriesByMod[modName] = queries
 }
 
-func (s autoCompleteSuggestions) sort() {
+func (s *autoCompleteSuggestions) sort() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
 	sortSuggestions := func(s []prompt.Suggest) {
 		sort.Slice(s, func(i, j int) bool {
 			return s[i].Text < s[j].Text


e-gineer:
This PR is no longer needed as the issue has already been fixed.

Bug #4711 (race condition in autoCompleteSuggestions.sort()) is a duplicate of bug #4716, which was already fixed in PR #4737 and merged to develop at commit 114ac22d.

The fix in develop already includes:
- sync.RWMutex field in autoCompleteSuggestions struct
- Lock/Unlock synchronization in sort() method
- Pointer receiver for sort() method

Closing as duplicate.

PR #4906, opened 2025-11-16
Fix #4761: Add timeout to prevent deadlock in RefreshConnections double-lock mechanism

### LINKED ISSUE #4761: Potential deadlock in RefreshConnections double-lock mechanism
## Issue Description

The `RefreshConnections` function uses a double-lock mechanism (queueLock + executeLock) that could theoretically lead to deadlock or starvation in certain edge cases.

## Location

`pkg/connection/refresh_connections.go:34-51`

## Severity

**MEDIUM** - Potential issue that could cause system hang in edge cases

## Details

The current locking mechanism works as follows:

```go
// Line 34-51
if !queueLock.TryLock() {
    // someone has it - they will execute so we have nothing to do
    log.Printf("[INFO] another execution is already queued - returning")
    return &steampipeconfig.RefreshConnectionResult{}
}

log.Printf("[INFO] acquired refreshQueueLock, try to acquire refreshExecuteLock")

// so we have the queue lock, now wait on the execute lock
executeLock.Lock()
defer func() {
    executeLock.Unlock()
    log.Printf("[INFO] released refreshExecuteLock")
}()

// we have the execute-lock, release the queue-lock so someone else can queue
queueLock.Unlock()
```

## Potential Deadlock Scenario

1. Goroutine A: `queueLock.TryLock()` succeeds
2. Goroutine A: Blocks on `executeLock.Lock()` (waiting indefinitely)
3. Goroutine B: `queueLock.TryLock()` fails, returns immediately
4. If `executeLock` is held by a goroutine that crashes/hangs, Goroutine A is deadlocked
5. All subsequent goroutines will fail at `queueLock.TryLock()` and return

## Impact

- Potential system hang if executeLock is never released
- Starvation of refresh operations
- No timeout mechanism to recover from hung goroutines

## Current Mitigations

- The code uses `defer` to release locks (good practice)
- Panic recovery in RefreshConnections returns error result
- Multiple concurrent calls are handled gracefully under normal conditions

## Suggested Improvements

1. **Add context-based timeout:**
```go
ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
defer cancel()

// Use context-aware locking or timeout mechanism
```

2. **Add monitoring/logging for long lock holds:**
```go
lockAcquiredTime := time.Now()
defer func() {
    holdTime := time.Since(lockAcquiredTime)
    if holdTime > 30*time.Second {
        log.Printf("[WARN] executeLock held for %v", holdTime)
    }
}()
```

3. **Consider using channels instead of locks:**
```go
type refreshRequest struct {
    ctx context.Context
    resultChan chan *steampipeconfig.RefreshConnectionResult
}
```

## Test Coverage

Added test in `pkg/connection/connection_lifecycle_test.go`:
- `TestRefreshConnectionsLocksCanDeadlock` - Documents the potential deadlock scenario
- `TestRefreshConnectionsConcurrentCalls` - Tests concurrent access behavior
- `TestRefreshConnectionsQueueBehavior` - Tests queue lock behavior

## Found By

Wave 3 testing expansion - Task 8 (Connection Package Testing)

## Note

This is not a confirmed bug but an area of architectural concern identified during comprehensive testing. The current implementation works correctly under normal conditions.

e-gineer:
## Summary
The RefreshConnections function uses a double-lock mechanism (queueLock + executeLock) that could lead to indefinite blocking if the executeLock holder crashes or hangs. This fix adds a 5-minute timeout to prevent deadlock.

## Root Cause
The current implementation blocks indefinitely when acquiring executeLock:
1. Goroutine A acquires queueLock via TryLock()
2. Goroutine A blocks on executeLock.Lock() waiting indefinitely
3. If the executeLock holder crashes/hangs, Goroutine A is deadlocked
4. All subsequent goroutines fail at queueLock.TryLock() and return
5. No timeout mechanism exists to recover from this state

## Solution
Added a timeout mechanism using goroutines and channels:
- Acquire executeLock asynchronously in a goroutine
- Use select with time.After to implement a 5-minute timeout
- If timeout occurs, release queueLock and return an error
- Normal operation continues unchanged when locks are acquired promptly

This ensures the system can recover from hung lock holders instead of blocking forever.

## Testing
- Added test demonstrating the potential deadlock scenario
- Verified fix resolves the issue with -race flag
- All existing connection package tests continue to pass

Fixes #4761

### DIFF
--- a/pkg/connection/connection_lifecycle_test.go
+++ b/pkg/connection/connection_lifecycle_test.go
@@ -6,6 +6,8 @@ import (
 	"sync/atomic"
 	"testing"
 	"time"
+
+	"github.com/turbot/steampipe/v2/pkg/steampipeconfig"
 )
 
 // TestExemplarSchemaMapConcurrentAccess tests concurrent access to exemplarSchemaMap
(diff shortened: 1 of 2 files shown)


e-gineer:
## Analysis of the 5-Minute Timeout

I've analyzed the double-lock mechanism in `RefreshConnections` and have concerns about adding a 5-minute timeout.

### How the Double-Lock Mechanism Works

The current implementation uses two locks:

1. **queueLock** (TryLock): Prevents multiple queued refreshes
   - First caller gets the lock, subsequent callers fail immediately and return
   - At most ONE refresh can be queued waiting for executeLock
   
2. **executeLock** (blocking Lock): Protects the actual refresh operation
   - The queued refresh waits here for the active refresh to complete
   - This waiting is **intentional** - we want exactly one queued refresh

```go
// Goroutine A: Gets queueLock, waits on executeLock (queued)
// Goroutine B: Fails at queueLock, returns immediately 
// Goroutine C: Fails at queueLock, returns immediately
```

**Key insight**: Only the first queued caller waits. All others fail fast at the queueLock.

### Problems with the 5-Minute Timeout

**1. False Failures for Legitimate Operations**

Operations that can legitimately take > 5 minutes while holding executeLock:
- Starting many plugins (especially on slower systems)
- Large schema updates (many connections × many tables)
- Network delays downloading plugin binaries
- Rate limiter definition updates across many plugins
- Connection state table updates for large installations

A timeout causes the queued refresh to fail even though the active refresh is working correctly.

**2. Creates Ambiguous Failure State**

When the timeout fires:
- Active refresh continues running (still holds executeLock)
- Queued refresh returns error to caller
- System is now in unclear state: "Is there a problem or not?"

**3. Defer Blocks Already Handle Crashes**

The existing defer statements release locks on panic/crash:
```go
defer func() {
    executeLock.Unlock()
    log.Printf("[INFO] released refreshExecuteLock")
}()
```

If the goroutine crashes, the defer runs and releases the lock. This prevents true deadlock from crashes.

**4. Masks Real Bugs Instead of Fixing Them**

If the code truly hangs forever (infinite loop, deadlock, etc.), that's a bug in the refresh logic that should be fixed, not worked around with a timeout.

### Recommendation

**Remove the timeout** for these reasons:

1. The queuing is intentional - when a refresh is queued, it *wants* to wait for the active one to complete
2. Large installations can legitimately need > 5 minutes
3. Defer blocks already handle crash scenarios
4. True hangs indicate bugs that should be fixed at the source

If protection against hangs is absolutely required, consider:
- A much longer timeout (30+ minutes) with clear logging
- Monitoring/alerting on refresh duration instead
- Adding intermediate progress logging to help diagnose actual hangs

The current 5-minute timeout will cause false failures in production for users with large Steampipe installations.

e-gineer:
Closing this PR based on the analysis above. The 5-minute timeout would cause false failures for large installations where refresh operations legitimately take longer. The existing defer blocks already protect against crashes, and the queueing behavior is intentional, not a bug to work around.

PR #4908, opened 2025-11-16
Fix #4790: Race condition between StreamRow() and Close()

### LINKED ISSUE #4790: Race condition between StreamRow() and Close()

e-gineer:
## Summary
Fixes a race condition where `StreamRow()` could panic with "send on closed channel" when called concurrently with `Close()` on a `Result` object. The fix uses atomic coordination to track the closed state and gracefully handles concurrent access.

## Root Cause
The race condition occurred when:
1. `StreamRow()` is actively sending rows to the result channel
2. `Close()` closes the channel while `StreamRow()` is in progress
3. Any in-flight `StreamRow()` call attempting to send after the channel closes causes a panic

The previous implementation had no synchronization between these operations, allowing `Close()` to close the channel while `StreamRow()` was about to send.

## Solution
Implemented a two-layer defense:
1. **Atomic closed flag**: Added `atomic.Bool` to track whether `Close()` has been called. `StreamRow()` checks this flag before attempting to send, preventing most race conditions.
2. **Panic recovery**: Added `defer/recover` in `StreamRow()` to gracefully handle the tiny window where `Close()` could occur between the flag check and the channel send operation.

This approach avoids deadlocks by not holding locks during blocking channel operations, while using atomic operations for lightweight coordination. Rows sent after close are silently dropped, which is safe behavior for concurrent scenarios.

## Testing
- Added `TestResult_CloseAfterPartialRead` that demonstrates the race condition (test fails before fix, passes after)
- Test creates concurrent goroutines that stream rows and close the result
- Verified fix resolves the issue - test passes consistently across 10 runs
- All existing tests continue to pass

Fixes #4790

### DIFF
--- a/pkg/query/queryresult/result.go
+++ b/pkg/query/queryresult/result.go
@@ -2,14 +2,17 @@ package queryresult
 
 import (
 	"sync"
+	"sync/atomic"
 
 	"github.com/turbot/pipe-fittings/v2/queryresult"
 )
 
 // Result wraps queryresult.Result[TimingResultStream] with idempotent Close()
+// and synchronization to prevent race between StreamRow and Close
 type Result struct {
 	*queryresult.Result[TimingResultStream]
 	closeOnce sync.Once
+	closed    atomic.Bool
 }
 
 func NewResult(cols []*queryresult.ColumnDef) *Result {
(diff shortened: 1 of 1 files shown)


e-gineer:
Closing this PR after thorough analysis and testing.

## Issue

While the atomic approach functionally works (all tests pass with defer/recover catching panics), the race detector flags a data race on the channel itself:

```
WARNING: DATA RACE
Write at channel (Close closing the channel)
Previous read at channel (StreamRow sending to channel)
```

The race occurs because even though we check `closed.Load()` before sending, there's still a window where `Close()` can close the channel while `StreamRow()` is executing the send operation. The defer/recover handles the panic, but the race detector sees unsynchronized concurrent access to the channel's internal data structure.

## Why This Matters

- Go's race detector flags **any** unsynchronized concurrent memory access to shared memory
- Even with panic recovery, we have concurrent operations on the channel metadata
- This could theoretically cause issues on some architectures or with future Go runtime changes
- Race detector warnings make it harder to catch real races in testing

## Decision

The current RWMutex-based implementation (from PR #4874) is the better solution:
- ✅ Race detector clean - no warnings
- ✅ Proven safe and predictable
- ✅ Future-proof against Go runtime changes  
- ✅ Performance difference is negligible for this use case

The atomic approach trades proper synchronization for a small performance gain, but introduces race detector warnings that we'd have to ignore/suppress. Not worth the tradeoff for this critical code path.

Keeping the current mutex-based implementation.

PR #4947, opened 2026-02-27
Warn after login when env vars override saved credentials

### LINKED ISSUE #4946: steampipe login should warn when PIPES_TOKEN or PIPES_HOST env vars will override login credentials
## Summary

After a successful `steampipe login --pipes-host <host>`, the saved token can be silently overridden by environment variables, leading to confusing failures when subsequently running queries.

## Problem

The login command saves the token to a per-host file (`~/.pipes/internal/<host>.tptt`) and prints "Logged in as: ...". However, two environment variables can silently make the login ineffective:

1. **`PIPES_TOKEN`** — has higher precedence than the saved token (env var at step 3 vs saved token at step 1 in `setCloudTokenDefault`), so subsequent commands will use the env var token instead of the freshly saved one.
2. **`PIPES_HOST`** — if set to a different host than the `--pipes-host` used for login, subsequent commands will look for the token file under the wrong host name and also hit the wrong API endpoint.

The user gets no warning at login time and only discovers the problem when queries fail.

## Expected Behavior

After a successful login, `steampipe login` should warn if:
- `PIPES_TOKEN` env var is set (it will override the saved login token)
- `PIPES_HOST` env var is set to a value different from the host used for login (subsequent commands will target a different host)

## Steps to Reproduce

```bash
export PIPES_TOKEN=spt_old_token
steampipe login --pipes-host custom.pipes.host.com
# Login succeeds, prints "Logged in as: ..."
# No warning that PIPES_TOKEN will override the saved token
steampipe query --workspace-database org/workspace "select 1"
# Fails or uses wrong credentials
```

graza-io:
## Summary

After a successful `steampipe login`, the saved token can be silently overridden by environment variables, leading to confusing failures when running subsequent queries. This PR adds post-login warnings to alert users.

### Problem

In `setCloudTokenDefault` (`pkg/cmdconfig/cmd_hooks.go`), the token precedence order is:
1. Saved token file (`~/.pipes/internal/<host>.tptt`)
2. Default workspace profile `cloud_token`
3. **`PIPES_TOKEN` env var** — overrides the saved token
4. Explicit workspace profile

Similarly, if `PIPES_HOST` env var is set to a different host than the one used for `--pipes-host`, subsequent commands silently target the wrong host.

### Fix

Added `showLoginWarnings()` called after a successful login in `displayLoginMessage()`. It checks:
- **`PIPES_TOKEN`**: If set (even to empty string), warns that it will take precedence over the login token, with an `unset` command hint
- **`PIPES_HOST`**: If set to a value different from the login host (from `--pipes-host` flag or default), warns about the mismatch

### Changes

- `cmd/login.go` — Added `showLoginWarnings()` function, called from `displayLoginMessage()`
- `cmd/login_test.go` — Unit tests for all warning scenarios

### Unit tests (`cmd/login_test.go`)

| Test Case | PIPES_TOKEN | PIPES_HOST | Expected |
|-----------|------------|------------|----------|
| No env vars set | unset | unset | No warnings |
| PIPES_TOKEN set | `spt_some_token` | unset | Token warning |
| PIPES_HOST differs from login host | unset | `other.pipes.host.com` | Host warning |
| PIPES_HOST matches login host | unset | same as login | No warnings |
| Both set, host differs | set | `other.pipes.host.com` | Both warnings |
| Both set, host matches | set | same as login | Token warning only |
| PIPES_TOKEN set to empty string | `""` | unset | Token warning (env var is still "set") |

### Manual tests

- [ ] Run `export PIPES_TOKEN=foo && steampipe login` — verify warning about `PIPES_TOKEN` is shown
- [ ] Run `export PIPES_HOST=other.host && steampipe login --pipes-host pipes.turbot.com` — verify warning about `PIPES_HOST` mismatch is shown
- [ ] Run `steampipe login` with neither env var set — verify no warnings

Closes #4946

### DIFF
--- a/cmd/login_test.go
+++ b/cmd/login_test.go
@@ -1,7 +1,9 @@
 package cmd
 
 import (
+	"io"
 	"os"
+	"strings"
 	"testing"
 
 	"github.com/fatih/color"
@@ -12,8 +14,9 @@ import (
 
 func TestShowLoginWarnings(t *testing.T) {
 	// disable color output for consistent test assertions
+	prevNoColor := color.NoColor
 	color.NoColor = true
-	defer func() { color.NoColor = false }()
+	t.Cleanup(func() { color.NoColor = prevNoColor })
 
 	tests := []struct {
 		name           string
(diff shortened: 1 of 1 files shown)


pskrbasu:
Addressed all Copilot review comments:

1. **`color.NoColor` restore** — Now saves previous value and restores via `t.Cleanup`
2. **`os.Unsetenv` leaking** — Extracted `setOrUnsetEnv` helper that saves/restores via `t.Cleanup` for the unset case
3. **Custom substring helpers** — Replaced with `strings.Contains`
4. **`color.Error` global restore** — Now restored via `t.Cleanup`, and errors from `os.Pipe`/`io.ReadAll` are checked with `t.Fatalf`

PR #4981, opened 2026-04-21
[CI TEST - DO NOT MERGE] vanilla v2.1.x acceptance sanity check

kaidaguerre:
Testing whether cloud/snapshot acceptance tests fail on vanilla v2.1.x HEAD with no pipe-fittings bump, to diagnose whether #4980 failures are caused by the bump or by env/infra drift. Will close after CI completes.


kaidaguerre:
Investigation complete — vanilla v2.1.x HEAD fails cloud+snapshot identically to #4980, confirming failures are env/infra drift, not caused by the pipe-fittings bump.

PR #4993, opened 2026-05-19
Make embedded-PG migration version-aware and non-bricking

kaidaguerre:
## What

The embedded-database migration path used a bare string inequality (`de.Name() != constants.DatabaseVersion`) and unconditionally ran `pg_dump` + `pg_restore`, **hard-failing service start** when the restore failed. That is unsafe for a cross-major upgrade (e.g. 14 → 18), where `pg_restore` cannot load a lower-major dump into a higher-major server.

Detection is now a semantic three-way branch (`classifyPgMigration`, parsed with `Masterminds/semver/v3` — the existing steampipe convention, no new dependency):

- **same-major / minor** (e.g. 14.17 → 14.19): keep today's automatic dump+restore; a restore *failure* is now **non-fatal** (retain insurance dump, keep old data dir, warn honestly, service still starts) instead of bricking.
- **cross-major** (e.g. 14 → 18): **Option B** — take the insurance dump, do NOT auto-restore, keep the old data dir, start fresh, print an honest "not auto-migrated / here's how to restore" message. (Option B is the owner-decided policy; the old data dir is preserved, which also enables rollback to the prior version.)
- unparseable version → classified minor (never strands data on a routine bump).

`start_services.go` no longer hard-fails on a migration error (degraded to a warning). `noBackupWarning()` no longer hard-codes "Postgres 14.17 to Postgres 14.19" — replaced with version-aware messages.

## Risk

Ships on **PG14, no version flip** — this changes the migration/startup path all users hit on a minor upgrade. The minor path is preserved (still auto dump+restore); the only changed minor behaviour is that a restore *failure* no longer bricks the service — strictly safer. This is the branch worth the closest review.

## Testing

- `classifyPgMigration` unit tests (same-major / cross-major / unparseable) — 9/9.
- `go test ./...` — 21 ok / 0 FAIL (= develop baseline).
- `migration.bats` real PG14→PG14 upgrades (v1.0.3, v2.2.0) — green, including a new assertion that the insurance dump is retained (the sanctioned migration-test change).
- Cross-major 14→18 (Option B) proven on real PG14+PG18 binaries: service starts on PG18, old dir kept, real PG14 dump retained, honest message — no brick.
- Not yet run on this branch: the full 22-file acceptance suite (run on the develop baseline only); the formal cross-major `migration.bats` fixtures are a separate follow-up.

## Scope

No embedded-version change here. The cross-major Option-B *integration* fixtures and the version flip are handled separately.

### DIFF
--- a/pkg/db/db_local/backup.go
+++ b/pkg/db/db_local/backup.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Masterminds/semver/v3"
 	"github.com/shirou/gopsutil/process"
 	"github.com/turbot/go-kit/files"
 	"github.com/turbot/pipe-fittings/v2/app_specific"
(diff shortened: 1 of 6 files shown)


kaidaguerre:
Superseded by the consolidated `pg18` branch, which contains all commits from this branch (including the Increment-6 engine work) rebased onto current develop. Closing.

PR #5021, opened 2026-07-31
Do not let one bad connection stop all connections from syncing

### LINKED ISSUE #5020: One bad connection block silently stops ALL connections from syncing until restart
## Describe the bug

A single connection block with a duplicate name (or an invalid schema name) permanently stops Steampipe from syncing **any** connection config, while the service continues to look healthy.

The chain:

1. `loadConfig` (`pkg/steampipeconfig/load_config.go`) returns an error for the whole load when one connection block has a duplicate name or an invalid schema name.
2. `LoadConnectionConfig` re-parses the **entire** config folder, so that error recurs on every call.
3. `ConnectionWatcher.handleFileWatcherEvent` (`pkg/connection/connection_watcher.go`) returns early when the load reports an error — **before** calling `RefreshConnections`.

So for as long as the offending block exists, no file event ever reaches `RefreshConnections`. Connections written after that point never get a schema and never get a row in `steampipe_connection_state`; updated credentials in rewritten `.spc` files are never applied. The file watcher itself keeps running (it keeps registering new files), and existing connections keep answering queries, so from the outside the service looks fine.

The only log line is `[WARN] error loading updated connection config: ...`, which is not visible at the default log level — so there is no operator-visible signal at all.

## Steps to reproduce

Service mode, config folder with a number of connections that are being written/updated by an external process:

1. Start the service and confirm connections load (`select * from steampipe_internal.steampipe_connection_state`).
2. Add a `.spc` file declaring a connection name that already exists in another file.
3. Write any new connection config file, or update an existing one.

Expected: the new/updated connection is applied; the duplicate is reported.
Actual: nothing is applied — ever. `steampipe_connection_state` freezes at the moment the duplicate appeared, no schema is created for anything added afterwards, and rewritten credentials are never picked up. Restarting the service surfaces the real cause immediately:

```
Error: duplicate connection name: 'conn_a'
	(/home/steampipe/.steampipe/config/conn_a.spc:1)
	(/home/steampipe/.steampipe/config/conn_a_duplicate.spc:1)
```

I reproduced this on a service-mode deployment with a few hundred connections whose config files are generated by an automated process. The failure is easy to hit in that setup and hard to diagnose, because the symptom (new connections never becoming queryable, credentials going stale) points nowhere near the actual cause.

## Expected behavior

A problem with one connection should disable that connection, not every connection. The rest of the config should continue to load and sync, and the failure should be visible at default log levels.

## Proposed fix

Skip the offending connection with a warning rather than failing the whole load, and log the config-load failure at ERROR level so it is visible. PR to follow.

## Steampipe version

v2.4.4

kazvictor:
Closes #5020

## Problem

`loadConfig` fails the entire config load when a single connection block has a duplicate name or an invalid schema name. Since `LoadConnectionConfig` re-parses the whole config folder and `ConnectionWatcher.handleFileWatcherEvent` returns early on a load error, one bad block stops `RefreshConnections` from running **at all** — for every connection, on every subsequent file event, for as long as that block exists.

The result is a silent, permanent halt to connection syncing while the file watcher keeps running: connections written afterwards never get a schema or a `steampipe_connection_state` row, and updated credentials in rewritten config files are never applied. Existing connections keep serving queries so the service looks healthy, and the only log line is a `[WARN]` that is invisible at default log levels. Full detail in #5020.

## Changes

- **`pkg/steampipeconfig/load_config.go`** — a duplicate connection name or an invalid connection name now appends a warning diagnostic and skips that connection, instead of returning an error for the whole load. The remaining connections load and continue to sync.
- **`pkg/connection/connection_watcher.go`** — log a config-load failure at `ERROR` rather than `WARN`, and state the consequence in the message. This line names the file to fix and is the only signal that syncing has stopped, so it should be visible at default log levels.
- **`pkg/steampipeconfig/load_config_resilience_test.go`** — regression tests for both cases: the whole load must succeed, the good connections must be present, the bad one must be skipped, and a warning must be reported. Both tests fail without the fix and pass with it.

## Notes

- Behaviour for other config errors (malformed HCL, bad plugin blocks) is unchanged; this narrowly targets per-connection problems that today take down the entire config.
- The duplicate case keeps the **first** declaration and skips the later one, which matches the existing error message's framing of "existing" vs "new".
- `go build ./...` and `go vet ./pkg/steampipeconfig/` pass.

### DIFF
--- a/pkg/connection/connection_watcher.go
+++ b/pkg/connection/connection_watcher.go
@@ -75,8 +75,17 @@ func (w *ConnectionWatcher) handleFileWatcherEvent([]fsnotify.Event) {
 	if !errorsAndWarnings.Empty() {
 		w.pluginManager.SendPostgresErrorsAndWarningsNotification(ctx, errorsAndWarnings)
 		// if there was an error return
+		//
+		// NOTE: this aborts the refresh for EVERY connection, not just the one
+		// the config error relates to, and it will keep doing so on every
+		// subsequent file event for as long as the config folder fails to
+		// parse - so connections silently stop being created and credential
+		// updates stop being applied while the watcher itself keeps running.
+		// Log at ERROR level so this is visible at default log levels: it
+		// names the file to fix, and it is the only signal that syncing has
+		// stopped.
 		if errorsAndWarnings.GetError() != nil {
-			log.Printf("[WARN] error loading updated connection config: %v", errorsAndWarnings.GetError())
+			log.Printf("[ERROR] error loading updated connection config - NO connections will be refreshed until this is resolved: %v", errorsAndWarnings.GetError())
 			return
 		}
 	}
--- a/pkg/steampipeconfig/load_config.go
+++ b/pkg/steampipeconfig/load_config.go
@@ -288,12 +288,29 @@ func loadConfig(ctx context.Context, configFolder string, steampipeConfig *Steam
 			if moreDiags.HasErrors() {
 				continue
 			}
+			// NOTE: a problem with a SINGLE connection must not fail the whole
+			// config load. LoadConnectionConfig re-parses the entire config
+			// folder, and the connection watcher aborts the refresh when it
+			// returns an error - so failing here for one bad connection stops
+			// ALL connections from being synced (no schemas created, no
+			// credential updates applied) for as long as the offending block
+			// exists, while the file watcher itself keeps running. Skip the
+			// offending connection with a warning and keep the rest usable.
 			if existingConnection, alreadyThere := steampipeConfig.Connections[connection.Name]; alreadyThere {
-				err := getDuplicateConnectionError(existingConnection, connection)
-				return perror_helpers.NewErrorsAndWarning(err)
+				diags = append(diags, &hcl.Diagnostic{
+					Severity: hcl.DiagWarning,
+					Summary:  getDuplicateConnectionError(existingConnection, connection).Error(),
+					Subject:  hclhelpers.BlockRangePointer(block),
+				})
+				continue
 			}
 			if ok, errorMessage := db_common.IsSchemaNameValid(connection.Name); !ok {
-				return perror_helpers.NewErrorsAndWarning(sperr.New("invalid connection name: '%s' in '%s'. %s ", connection.Name, block.TypeRange.Filename, errorMessage))
+				diags = append(diags, &hcl.Diagnostic{
+					Severity: hcl.DiagWarning,
+					Summary:  fmt.Sprintf("invalid connection name: '%s' in '%s'. %s ", connection.Name, block.TypeRange.Filename, errorMessage),
+					Subject:  hclhelpers.BlockRangePointer(block),
+				})
+				continue
 			}
 			steampipeConfig.Connections[connection.Name] = connection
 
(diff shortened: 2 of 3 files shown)


kaidaguerre:
Thanks for looking at this — the diagnosis is right, and skipping the bad connection with a warning is the behaviour we want. Raising that log line to ERROR is a good call too.

One change needed here before we merge: `tests/acceptance/test_files/connection_config.bats:62` is named "steampipe should return an error for duplicate connection name" and asserts the old behaviour. It only matches on output text, so it will still pass against your change while asserting the opposite of its name. It should expect the query to succeed and the duplicate to be reported as a warning. (Ignore the two red checks from July — those runs expired waiting on our approval rather than failing; your next push will re-queue them.)

One thing this doesn't cover: the freeze itself is the early return in `handleFileWatcherEvent`, which fires on any config-load error — so a malformed `.spc`, a duplicate plugin block label or a bad options block still stops all syncing with the same invisible symptom you described. That wants the watcher to refresh whatever did load rather than bailing out. I've raised #5039 for it — feel free to pick that one up as well if you'd like.

kazvictor:
Thanks! The test is updated in 8522ee0. it now asserts the query succeeds and the duplicate is reported as a warning, and it is renamed to match what it actually checks.

kaidaguerre:
Sorry this took a while to get to — CI on fork PRs needs a maintainer to approve each run, and your push sat waiting for that. It's run now.

Lint passes, but the updated duplicate-name test fails in the acceptance suite (`connection_config`). The duplicates are reported as warnings as intended, but the query then fails during database startup:

```
not ok 5 steampipe should warn on duplicate connection name and still run the query
status : 250
  Warning: duplicate connection name: 'chaos'
  ...
  Warning: duplicate connection name: 'chaos3'
  Error: failed to initialise functions: conn closed
```

The rest of `connection_config` passes, and the same suite is green on `develop`, so it looks tied to the change rather than a flake. The error comes from `pkg/db/db_local/internal.go:159`.

Could you take a look and run it locally before the next push?

```
go build -o steampipe && export PATH=$PWD:$PATH
tests/acceptance/run-local.sh connection_config.bats
```

Full log: https://github.com/turbot/steampipe/actions/runs/34983080266

kazvictor:
Ran it locally, same failure on linux. Dug into the database logs and it's not the CLI side; the FDW is what's dying.

With duplicates downgraded to warnings the service start now goes ahead, and the `IMPORT FOREIGN SCHEMA` in setupInternal is the first statement that loads the FDW into a backend. The FDW loads the connection config itself (fdw.go:70 → `LoadConnectionConfig` in hub_remote.go) using the steampipe version pinned in its own go.mod (v2.2.0-rc.0), which still hard-errors on duplicate names, and `init()` panics on that. The panic kills the backend and the CLI just sees the closed connection:

```
[WARN]  hub: LoadConnectionConfig failed {duplicate connection name: 'chaos' ...}
panic: duplicate connection name: 'chaos'
main.init.0() fdw.go:70
LOG:  server process (PID ...) was terminated by signal 6: Aborted
DETAIL:  Failed process was running: IMPORT FOREIGN SCHEMA "steampipe_internal" FROM SERVER steampipe INTO steampipe_internal;
```

On develop the CLI rejects the config before the service ever starts, so the FDW never sees it; that's why the suite is green there.

Until the FDW vendors a loader that tolerates this, any command that starts the service hits the same crash when the config has duplicate (or invalid-name) connections. Pushed 6ffabe0b: test 5 now asserts the warning plus a successful `plugin install`, which covers the loader path without starting the db. The query assertion can go back in once the FDW catches up.

